PR #27, opened 2024-07-23
Add initDevices lifecycle method

Regaddi:
This adds a new `initDevices()` lifecycle method, which can be optionally called before `start()` from a voice client instance.
`initDevices()` is essentially a wrapper around `startCamera()` in the Daily transport layer, which allows developers to tie the initialization of devices and audio contexts to a user gesture, as required by browsers.
This adds two new transport states: `"initializing"` and `"initialized"`.
`initDevices()` is a no-op when the transport state is not `"idle"`.


jptaylor (approved):
LGTM

PR #31, opened 2024-07-24
remove unused property startMicMuted

Regaddi:


jptaylor (approved):

jptaylor (commented):
@Regaddi we can remove this property, although eventually I think we need this in the SDK. In the current demos, you can enable a mic but join muted. Joining muted is different from removing the device source entirely.

Perhaps better if we implement that, vs. remove these properties?

Regaddi:
The way `enableMic` is currently implemented is essentially a wrapper around `localAudio` and `setLocalAudio`.
Perhaps the name isn't necessarily reflective of what it's actually doing, but it seemed to me like `enableMic` and `startMicMuted` are kind of the same thing. I don't think we'll need to handle mic permissions (as in daily-js `canSend` permissions) as part of this SDK, if that's something you envisioned for `enableMic` instead.

PR #37, opened 2024-07-25
Prevent state mutation when new state is the same

jptaylor:
Some workflows may signal a ready state more than once. This check avoids a mutation of the property when the values are the same.


LewisWolfgang (approved):
lgtm

PR #42, opened 2024-08-05
change message label from rtvi to  rtvi-ai

kwindla:
Label change to align with update on the Pipecat side.

There's more work to do here on the various message formats. But maybe makes sense to merge this, now.

Tested with:

```
voiceClient.say("What a beautiful day.")
```


jptaylor (approved):
lgtm

PR #53, opened 2024-08-18
version 0.1.4 patch

jptaylor:


rtvi-client-react/src/VoiceClientVideo.tsx
```diff
@@ -0,0 +1,192 @@
...
+}
+
+interface Props
+  extends Omit<React.VideoHTMLAttributes<HTMLVideoElement>, "onResize"> {
+  participant: "local" | "bot";
+
+  /**
+   * For local user-facing camera streams, we'll automatically mirror the video.
+   */
+  automirror?: boolean;
```
markbackman, line 18: Defined but not used.

markbackman (approved):
Generally looks good! There are two unused props, but I think that's OK for now. We can fix that up later.

PR #57, opened 2024-08-27
Fixing couple of warnings and making sandbox work directly with staging.

filipi87:
- Allowing the sandbox connect directly to the server using daily api key from the environment.
- Removing daily-js dependency that should not be needed by rtvi-client-js
- Fixed some lint warnings

### DIFF
--- a/rtvi-client-js-daily/src/transport.ts
+++ b/rtvi-client-js-daily/src/transport.ts
@@ -28,12 +28,8 @@ export interface DailyTransportAuthBundle {
 }
 
 export class DailyTransport extends Transport {
-  protected _state: TransportState = "idle";
-
   private _daily: DailyCall;
   private _botId: string = "";
-  private _expiry: number | undefined = undefined;
-
   private _selectedCam: MediaDeviceInfo | Record<string, never> = {};
   private _selectedMic: MediaDeviceInfo | Record<string, never> = {};
 
@@ -61,7 +57,7 @@ export class DailyTransport extends Transport {
     return this._state;
   }
 
-  private set state(state: TransportState) {
+  set state(state: TransportState) {
     if (this._state === state) return;
 
     this._state = state;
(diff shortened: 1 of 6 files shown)


rtvi-client-js-daily/src/transport.ts
```diff
@@ -28,12 +28,8 @@ export interface DailyTransportAuthBundle {
 }
 
 export class DailyTransport extends Transport {
-  protected _state: TransportState = "idle";
```
filipi87, line 31: I have removed this variable from here, because it is already declared inside the abstract class `Transport`.

rtvi-client-js/src/core.ts
```diff
@@ -313,6 +313,7 @@ export abstract class Client extends (EventEmitter as new () => TypedEmitter<Voi
               this._abortController!
             );
           } else {
+            // @ts-ignore
```
filipi87, line 316: I keep seing this error in my IDE:
```
Argument type {mode: "cors", headers: {'Content-Type': string}, method: string, body: string, signal: AbortSignal | null} is not assignable to parameter type RequestInit | undefined 
```

So, I have added it just as a workaround to stop it while I was looking at the other warnings. But I was not supposed to have commited this one. 
I need to figure out why is it complaining, because it seems that everything works just fine.

rtvi-client-js/src/core.ts
```diff
@@ -492,7 +493,7 @@ export abstract class Client extends (EventEmitter as new () => TypedEmitter<Voi
    * @param config - VoiceClientConfigOption[] partial object with the new configuration
    * @returns VoiceClientConfigOption[] - Updated configuration
    */
-  protected set config(config: VoiceClientConfigOption[]) {
+  public set config(config: VoiceClientConfigOption[]) {
```
filipi87, line 496: This one, I was receiving a warning, regarding different access modifiers for the `get` and `set`. 

But that is probably an issue from my IDE, I was just looking at this PR, and it seems it is supposed to work without any kind of warning: 
https://github.com/microsoft/TypeScript/pull/42425

Set accessors may be less visible than their corresponding get. 

So we should remove this change for sure.

filipi87:
Closing this PR, all the relevant changes will be included at this branch: `jpt/0.1.7`

PR #60, opened 2024-08-30
Implement screen track capabilities

Regaddi:
This implements screen sharing capabilities for the local participant.

**New APIs (on VoiceClient and Transport):**

- startScreenShare(): void
- stopScreenShare(): void
- isSharingScreen(): boolean

**New events:**

- onScreenTrackStarted(track: MediaStreamTrack, participant: Participant)
- onScreenTrackStopped(track: MediaStreamTrack, participant: Participant)
- onScreenShareError(errorMessage: string)

**React updates:**

- `useVoiceClientMediaTrack` now supports `screenAudio` and `screenVideo`
- `VoiceClientVideo` now accepts an optional `trackType` prop, which can be `"video" | "screenVideo"` and defaults to `"video"`

This is minimalistic support without detailed controls over the screen share. We can add configurable options for getDisplayMedia later, if we need to. For now calling `startScreenShare()` will trigger the default screen share prompt.

### DIFF
--- a/rtvi-client-js-daily/src/transport.ts
+++ b/rtvi-client-js-daily/src/transport.ts
@@ -3,6 +3,7 @@ import Daily, {
   DailyEventObjectAppMessage,
   DailyEventObjectAvailableDevicesUpdated,
   DailyEventObjectLocalAudioLevel,
+  DailyEventObjectNonFatalError,
   DailyEventObjectParticipant,
   DailyEventObjectParticipantLeft,
   DailyEventObjectRemoteParticipantsAudioLevel,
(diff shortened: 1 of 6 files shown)


jptaylor:
Holding on this PR until RTVI supports it on the bot side

KrishnaPG:
Wondering what is the status of this? The change log mentions it as available https://github.com/pipecat-ai/pipecat-client-web/blob/1b6891e749d1248e413135ab74d2cb01fe6d6145/CHANGELOG.md?plain=1#L37 but looks like this is still pending. Can this be pushed please?

Regaddi:
Ah, these changes made it in on separate branches. Closing now! It's all there already 😌

PR #69, opened 2024-09-14
RTVI 0.2.0

jptaylor:
RTVI 0.2.0 removes client-side configuration, ensuring that state management is handled exclusively by the bot or the developer’s application logic. Clients no longer maintain an internal config array that can be modified outside of a ready state. Developers who require stateful configuration before a session starts should implement it independently.

This change reinforces a key design principle of the RTVI standard: the bot should always be the single source of truth for configuration, and RTVI clients should remain stateless.

Additionally, this release expands action capabilities, enabling disconnected action dispatch, and renames key classes and types from` VoiceClientX` to `RTVIClientX`. Where possible, we have left deprecated aliases to maintain backward compatibility.


CHANGELOG.md
```diff
@@ -5,6 +5,59 @@ All notable changes to **RTVI Client Web** will be documented in this file.
...
+### Changed
+
+- `start()` has been renamed to `connect()`.
+- Client no longer expects a `services` map as a constructor param (note: remains in place but flagged as deprecated.) If you want to pass a services map to your endpoint, please use `params`.
+- `customHeaders` has been renamed to `headers`.
+- Config getter and setter methods (`getConfig` and `updateConfig`) are only supported at runtime. 
+- `updateConfig` promise is typed to `Promise<VoiceMessage>` (previously `unknown` to support offline updates.)
+- `getConfig` promise is typed to `Promise<VoiceClientOptions[]>` (previously `unknown` to support offline updates.)
+- `services` getter and setter methods have been deprecated.
+- `getServiceOptionsFromConfig`, `getServiceOptionValueFromConfig`, `setConfigOptions` and `setServiceOptionInConfig` are now async to support `getConfig` at runtime and accept an optional `config` param for working with local config arrays.
```
marcus-daily, line 32: Better to just always require the `config` parameter, and never fetch the config inside the method? (and therefore make it non-async?) Otherwise I can see people calling this multiple times to fetch several values, and causing a round trip each time.

It would be nice to separate these from the `RTVIClient` entirely and have them be utility functions in a separate namespace.

CHANGELOG.md
```diff
@@ -5,6 +5,59 @@ All notable changes to **RTVI Client Web** will be documented in this file.
...
+- `registerHelper` no longer checks for a registered service and instead relies on string matching.
+- LLM Helper `getContext()` now accepts optional `config` param for working with local configs.
+- `customAuthHandler` updated to receive `startParams` as second dependency.
+- jest tests updated to reflect changes.
+- `VoiceClientOptions` is now `RTVIClientOptions`.
+- `VoiceClientConfigOption` is now `RTVICClientConfigOption`.
```
marcus-daily, line 38: Typo (RTVICClient)

rtvi-client-js/src/client.ts
```diff
@@ -0,0 +1,1031 @@
...
+
+  /**
+   * Service configuration options for services and further customization
+   * @deprecated Use params.config instead
+   */
+  config?: VoiceClientConfigOption[];
+
+  /**
+   * Custom HTTP headers to be send with the POST request to baseUrl
+   * @deprecated Use startHeaders instead
```
marcus-daily, line 130: Should this be `headers` rather than `startHeaders`?

rtvi-client-js/src/client.ts
```diff
@@ -0,0 +1,1031 @@
...
+
+export type RTVIClientParams = {
+  baseUrl: string;
+} & Partial<{
+  headers?: Headers;
+  endpoints: Record<RTVIURLEndpoints, string>;
+  requestData?: object;
+  config?: RTVIClientConfigOption[];
+}> & {
+    [key: string]: unknown;
```
marcus-daily, line 51: What are the extra optional fields for here? This pattern isn't possible on mobile.

rtvi-client-js/src/client.ts
```diff
@@ -0,0 +1,1031 @@
...
+
+  /**
+   * Custom HTTP headers to be send with the POST request to baseUrl
+   * @deprecated Use startHeaders instead
+   */
+  customHeaders?: { [key: string]: string };
+
+  /**
+   * Custom request parameters to send with the POST request to baseUrl
+   * @deprecated Use params instead
```
marcus-daily, line 136: `params` -> `params.requestData`?

rtvi-client-js/src/messages.ts
```diff
@@ -159,17 +181,66 @@ export class MessageDispatcher {
...
+        message: action,
+        timestamp: Date.now(),
+        resolve,
+        reject,
+      });
+    });
+
+    console.debug("[MessageDispatcher] action", action);
+
+    if (this._client.connected) {
```
marcus-daily, line 207: Race condition -- if the transport disconnects while the user is sending an action, that action will be sent to the "single turn" endpoint, rather than the intended "connected" endpoint.

marcus-daily: There's not really much reason to combine these into a single `action()` API, vs having `action()` and `actionSingleTurn()`, for example, which would allow the user to specify exactly what they want, without resorting to race-prone state checks.

markbackman (approved):
LGTM!

PR #72, opened 2024-10-21
RTVI 0.2.1

jptaylor:
- Daily package moved to `@daily-co/realtime-ai-daily`
- Fixed deprecation warning for services map


rtvi-client-js/src/client.ts
```diff
@@ -165,21 +165,19 @@ export type RTVIEventCallbacks = Partial<{
...
-  onBotLlmStarted: (participant: Participant) => void;
-  onBotLlmStopped: (participant: Participant) => void;
-
-  onBotTtsText: (text: TTSTextData) => void;
-  onBotTtsStarted: (participant: Participant) => void;
-  onBotTtsStopped: (participant: Participant) => void;
+  onBotTranscript: (data: BotLLMTextData) => void;
+  onBotLlmText: (data: BotLLMTextData) => void;
+  onBotLlmStarted: () => void;
+  onBotLlmStopped: () => void;
```
aconchillo, line 177: What is the `participant` in `onBotStartedSPeaking`? Should that be removed as well?

jptaylor: @aconchillo you might be right, but I thought we did receive a participant object there. Let me check and update accordingly.

jptaylor: Confirmed these are not implemented. Removed.

markbackman (approved):
LGTM!

PR #76, opened 2024-11-08
Add disconnectBot message and helper

Regaddi:
This is the client-side part of the new `disconnect-bot` RTVI message.


markbackman:
I think this looks good. I'm just using for feedback on naming from https://github.com/pipecat-ai/pipecat/pull/697.

PR #77, opened 2024-11-12
Logger and log levels

jptaylor:
Implemented package-wide logger so that custom log levels can be set for RTVI (without relying on global console log levels)

```typescript
import {logger, LogLevel} from "realtime-ai"

logger.setLevel(LogLevel.DEBUG); // default
logger.setLevel(LogLevel.INFO);
logger.setLevel(LogLevel.WARN);
logger.setLevel(LogLevel.ERROR);
logger.setLevel(LogLevel.NONE);
```

Alternatively, set on the client object itself:

```typescript
rtviClient.setLogLevel(LogLevel.NONE);
```


markbackman (approved):
LGTM

PR #80, opened 2024-11-25
adding a getter and types for AuthBundle

David-Botos:


mattieruth:
Apologies for silence on this. And thank you for taking the initiative of providing a PR instead of just posting an issue!

The authBundle type is dependent on the transport type, and may not even be used (see the [OpenAIRealTimeWebRTCTransport](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/openai-realtime-webrtc-transport/src/OpenAIRealTimeWebRTCTransport.ts#L182) as an example. So I don't think we can properly type it on the Client. The type you have spelled out in this PR  is specific to the [DailyTransport and is typed there](https://github.com/pipecat-ai/pipecat-client-web-transports/blob/main/transports/daily/src/transport.ts#L31-L34).

Separately, since the AuthBundle is transport-specific, I think it makes sense to add the getter to the transport instead of the client. I'd like to understand the use case your solving for to make sure that a getter for the AuthBundle is what is really wanted (vs. getters for the room_url/token that is used). Once I understand the need, I will close this ticket and open a new one on the transports repo.

PR #81, opened 2024-11-26
Update RTVI client web for Pipecat

markbackman:
- Rename the directories
- Update the READMEs
- Add some header images
- Add CONTRIBUTING to the top level directory
- Add LICENSE file
- Add copyright header to all files

Proposed repo name: pipecat-client-web.

We'll refer to it as Pipecat Client Web or Pipecat Web when we refer to Pipecat JS or React collectively. And, we'll refer to it as Pipecat JS or Pipecat React when addressing the SDK directly. Note: we're still maintaining the realtime-ai npm name for packages.

I'm not planning to change the names away from RTVI. That breaking change seems like something we should avoid. WDYT?

Note: the images in the READMEs will be temporarily broken until we rename the repo.


package.json
```diff
@@ -1,11 +1,11 @@
 {
   "private": true,
-  "name": "rtvi-client-web",
+  "name": "pipecat-client-web",
   "version": "0.2.2",
   "workspaces": [
-    "rtvi-sandbox",
-    "rtvi-client-js",
-    "rtvi-client-react"
+    "pipecat-sandbox",
```
Regaddi, line 6: The sandbox doesn't exist anymore and can be removed here.

markbackman: Thanks!

README.md
```diff
@@ -1,14 +1,20 @@
-# Real-Time Voice Inference Web SDK
+<h1><div align="center">
+ <img alt="pipecat client web" width="500px" height="auto" src="https://raw.githubusercontent.com/pipecat-ai/pipecat-client-web/main/pipecat-flows.png">
```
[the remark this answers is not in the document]
markbackman: Yes, thanks for catching that typo.

PR #91, opened 2024-12-11
baseUrl and endpoints are now optional

jptaylor:
- [ ] Add additional tests
- [ ] Error relevance when baseUrl is not set but an auth bundle is required

### DIFF
--- a/client-js/src/client.ts
+++ b/client-js/src/client.ts
@@ -49,7 +49,7 @@ const defaultEndpoints: Record<RTVIURLEndpoints, string> = {
 };
 
 export type RTVIClientParams = {
-  baseUrl: string;
+  baseUrl?: string;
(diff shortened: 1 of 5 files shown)


client-js/src/client.ts
```diff
@@ -49,7 +49,7 @@ const defaultEndpoints: Record<RTVIURLEndpoints, string> = {
 };
 
 export type RTVIClientParams = {
-  baseUrl: string;
+  baseUrl?: string;
 } & Partial<{
   headers?: Headers;
   endpoints: Partial<Record<RTVIURLEndpoints, string>>;
```
mattieruth, line 55: should endpoints also be optional?

mattieruth (commented):
i think this makes sense and is a fine change.

PR #95, opened 2025-02-27
Added support for a server to client RTVI message

markbackman:
WIP


filipi87 (approved):
Yep, makes a lot of sense to make things easier.

PR #101, opened 2025-03-18
Improve types for llm helpers for easier parsing downstream and to ma…

mattieruth:
…ke it clearer what the options are


jptaylor (approved):
LGTM

PR #105, opened 2025-04-03
update transport wrapper to disallow calling initDevices directly

mattieruth:


client-js/src/transport.ts
```diff
@@ -109,6 +109,9 @@ export class TransportWrapper {
             case "initialize":
               errMsg = `Calls to initialize() are disabled and used internally by the RTVIClient`;
               break;
+            case "initDevices":
+              errMsg = `Calls to initDevices() are disabled and used internally by the RTVIClient`;
```
kompfner, line 113: `RTVIClient.initDevices()` is a thing that the user could call directly, right, to turn on local devices before connecting? If so, did we want to here say something along the lines of "Please use RTVIClient.initDevices() instead, or let RTVIClient call it automatically for you"

mattieruth: i thought my message was that, but i guess i copy/pasted the wrong one 🤦‍♀️

PR #106, opened 2025-04-07
copy/pasta :facepalm

mattieruth:


client-js/src/transport.ts
```diff
@@ -110,7 +110,7 @@ export class TransportWrapper {
               errMsg = `Calls to initialize() are disabled and used internally by the RTVIClient`;
               break;
             case "initDevices":
-              errMsg = `Calls to connect() are disabled. Please use RTVIClient.connect()`;
+              errMsg = `Calls to initDevices() are disabled. Please use RTVIClient.connect()`;
```
kompfner, line 113: Wait, it seems like my feedback from the last PR might not have gotten across. I was asking whether the user could legitimately want to call `RTVIClient.initDevices()` directly rather than have it automatically called through `connect()`, and—if so—whether that should be accounted for in this message.

kompfner (approved):

client-js/src/transport.ts
mattieruth: blech. i _still_ had copy/paste cruft. :embarrassed. I went ahead and tried to improve the wording of all the denied calls. Note: I decided _not_ to mention `connect()` as an option for `initDevices()` because it felt like TMI. If someone is trying to call `Transport.initDevices()`, then most likely they are indeed wanting to initialize the devices prior to connecting. And if they aren't, then it doesn't really matter whether they call it or `connect()` calls it.

kompfner (approved):

PR #107, opened 2025-04-08
automatically init devices from useRTVIClientMediaDevices

Regaddi:
Oftentimes, when conditionally rendering device-sensitive components, such as device pickers, they are prone to missing out RTVI events, specifically `AvailableXsUpdated` and `XUpdated` (X = Cam | Mic | Speaker).

To prevent those components reading from stale state, this PR adds a `useEffect` to lazily initialize device states.

### DIFF
--- a/client-react/src/useRTVIClientMediaDevices.ts
+++ b/client-react/src/useRTVIClientMediaDevices.ts
@@ -1,7 +1,7 @@
 import { RTVIEvent } from "@pipecat-ai/client-js";
 import { atom, useAtomValue } from "jotai";
 import { useAtomCallback } from "jotai/utils";
-import { useCallback } from "react";
+import { useCallback, useEffect } from "react";
 import { useRTVIClient } from "./useRTVIClient";
 import { useRTVIClientEvent } from "./useRTVIClientEvent";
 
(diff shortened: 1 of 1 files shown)


mattieruth (approved):
TBH, I don't _fully_ follow the issue and how this resolves it, but it seems fine. Is the idea to wrap `initDevices()` and have the app call this new `initDevices()` instead of the client directly?

Regaddi:
Let me try to explain:

When you build a Pipecat Client App, there's a likeliness that you end up having components to control your devices, however these might not be rendered immediately, like only when you open a "Device settings" modal or alike.
In those scenarios there wouldn't be any instance of `useRTVIClientMediaDevices`. At the time the session is initiated, we emit events (`AvailableCamsUpdated`, `AvailableMicsUpdated`, etc). If there's no instance of `useRTVIClientMediaDevices` rendered at that point in time, the React state remains empty, as if no devices are available, because the events weren't consumed.
One resolution to the problem is to always render a top-level instance of `useRTVIClientMediaDevices`, right under the `RTVIClientProvider`.
This PR implements a different solution: instead of having people change their code, in case they're rendering device controls conditionally, we'll automatically update the devices state once a component calling `useRTVIClientMediaDevices` is mounted. This makes sure, that device states are always up to date, independent of events.
A third option to solve this problem would be a top-level wrapper component that manages the state internally, instead of `useRTVIClientMediaDevices`. We followed this approach in Daily React (e.g. [DailyDevices](https://github.com/daily-co/daily-react/blob/main/src/DailyDevices.tsx)), but it adds component layers to the component stack. We might migrate to that approach, if we like, as it doesn't require code changes.

PR #111, opened 2025-05-06
add optional barCount prop to VoiceVisualizer component

Regaddi:
This adds a new optional `barCount` prop to `VoiceVisualizer` to allow for more customization of voice visualization.
I also changed the frequency band calculation to use the [Mel scale](https://en.wikipedia.org/wiki/Mel_scale) in order to calculate meaningful frequency bands for visualizing.
With large numbers of bars (60+) the algorithm will start to introduce empty bars, which is due to logarithmic nature of sound.

CleanShot 2025-05-06 at 16 16 10@2x

CleanShot 2025-05-06 at 16 29 51@2x

CleanShot 2025-05-06 at 16 30 32@2x


markbackman (approved):
Nice!

PR #117, opened 2025-06-04
Pipecat Client 1.0.0 🎉 

mattieruth:
🎉  Introducing Version 1.0.0 of the Javascript and React Pipecat Client APIs!

⚠️ THIS IS A BREAKING CHANGE: This PR introduces a major refactor of the RTVIClient, now called PipecatClient and all the client classes.

The client itself has been simplified to allow more flexibility and less built-in assumptions about how the client-server startup flow works.

The biggest changes are:

1. Name Changes

The `RTVIClient` is now the `PipecatClient` and many other types have been updated to match. However, the messaging protocol between client<->server remains identified as RTVI (Real Time Voice Inference) so as to define a generic AI client<->server messaging interface. The folder structure has been updated to represent this separation as well and all RTVI protocol code is now conveniently organized into its own `rtvi` module.

2. Pipeline configuration has been removed

Due to security concerns, the RTVI protocol has been updated to remove support for initial setup and dynamic updates to the pipeline service configurations and llm context. These messages could have allowed for malicious takeover of backend services. Pipeline setups that require client-side opinions should be handled using custom endpoints and client-server messaging (which has been made easier).

3. Actions and Helpers have been removed

Like configuration, the "actions" interface has been removed from the RTVI protocol in favor of the more simplistic approach of client-server messages and developer-defined payloads. This changes should make this functionality easier to understand and achieve. Some previously defined actions have been preserved and are now top-level RTVI messages, primarily, `llm-function-call`, and `bot-llm-search-response`. 

4. Authentication

The authentication logic is now optional and has moved to the `connect()` step for more direct and dynamic connection-time configuration. `connect()` now takes an optional set of parameters for defining connection-time information (like room/token in the case of a `DailyTransport` or a url in the case of a `SmallWebRTCTransport`). This allows a client to connect->disconnect repeatedly with different auth params.  Same as before, the `PipecatClient.connect()` wraps the Transport's `connect()` function and will pass the params along. For convenience, the `PipecatClient` `connect()` function allows an alternative parameter type, allowing you to pass in a url/headers/data for fetching the params the transport expects from a REST endpoint.

### DIFF
--- a/client-js/client/rest_helpers.ts
+++ b/client-js/client/rest_helpers.ts
@@ -2,10 +2,18 @@ import * as RTVIErrors from "../rtvi/errors";
 import { logger } from "./logger";
 import { TransportConnectionParams } from "./transport";
 
+type Serializable =
+  | string
+  | number
+  | boolean
+  | null
+  | Serializable[]
+  | { [key: number | string]: Serializable };
+
 export interface ConnectionEndpoint {
   endpoint: string;
   headers?: Headers;
-  requestData?: object;
+  requestData?: Serializable;
   timeout?: number;
 }
 


client-js/client/rest_helpers.ts
```diff
@@ -0,0 +1,88 @@
+import * as RTVIErrors from "../rtvi/errors";
+import { logger } from "./logger";
+import { TransportConnectionParams } from "./transport";
+
+export interface ConnectionEndpoint {
+  endpoint: string;
+  headers?: Headers;
+  requestData?: object;
```
Regaddi, line 8: `object` is a fairly open type, as it allows anything where `typeof` is `object`, so even things like `new Date()`, `new RegExp()`, `new WeakMap()` etc.
Do we want to narrow this a bit down? Or is this intentionally open, because the requestData can be literally anything?

client-js/client/transport.ts
```diff
@@ -39,29 +22,69 @@ export type Tracks = {
...
+   * Establishes a connection with the remote server. This is the main entry
+   * point for the transport to start sending and receiving media and messages.
+   * This is called from PipecatClient.connect() and should not be called directly.
+   * @param abortController
+   */
+  abstract _connect(connectParams?: TransportConnectionParams): Promise<void>;
+  /**
+   * Disconnects the transport from the remote server. This is called from
+   * PipecatClient.disconnect() and should not be called directly.
+   */
```
Regaddi, line 75: The other comments above are on the abstract methods. Should this move to `abstract _disconnect()`, too?

client-js/rtvi/events.ts
```diff
@@ -4,141 +4,152 @@
...
+
+  /** server messaging */
+  ServerMessage = "serverMessage",
+  ServerResponse = "serverResponse",
+  MessageError = "messageError",
+
+  /** service events */
+  Metrics = "metrics",
+
+  // vad events
```
Regaddi, line 39: nit: Perhaps streamline the comments to have the same format (`/** ... */` vs. `// ...`)

client-js/client/logger.ts
```diff
@@ -54,6 +54,6 @@ class Logger {
   }
 }
 
-export const logger = Logger.getInstance();
+export let logger = Logger.getInstance();
```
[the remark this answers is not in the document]
mattieruth: good catch. i don't think this was an intentional change.

client-js/rtvi/events.ts
mattieruth: ha. this was intentional but probably only made sense in my head. /** ... */ denote a larger class of event types and i was using // to denote sort of a sub-class of events under those.  So in this case, vad events are still service events, just related to the vad service. do you have thoughts on a better way to do this that's not obnoxious?

client-js/client/client.ts
```diff
@@ -0,0 +1,707 @@
...
+  /**
+   * Initialize local media devices
+   */
+  public async initDevices() {
+    logger.debug("[Pipecat Client] Initializing devices...");
+    await this._transport.initDevices();
+  }
+
+  /**
+   * The `connect` function in TypeScript establishes a transport session and
```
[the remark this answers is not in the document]
mattieruth: i don't either. i'm going to blame copilot autocomplete on this one 😂

client-js/client/client.ts
```diff
@@ -0,0 +1,707 @@
...
+   */
+  @transportReady
+  public async sendClientRequest(
+    msgType: string,
+    data: unknown,
+    timeout?: number
+  ): Promise<unknown> {
+    return new Promise((resolve, reject) => {
+      const msgData: ClientMessageData = { t: msgType, d: data };
+      this._messageDispatcher
```
kompfner, line 551: why not `await` syntax instead of `.then`? seems like you prefer `await` in `appendToContext()`

client-js/client/client.ts
```diff
@@ -0,0 +1,707 @@
...
+  protected handleMessage(ev: RTVIMessage): void {
+    logger.debug("[RTVI Message]", ev);
+
+    switch (ev.type) {
+      case RTVIMessageType.BOT_READY: {
+        const data = ev.data as BotReadyData;
+        const botVersion = data.version.split(".").map(Number);
+        if (botVersion[0] < 1) {
+          logger.warn(
+            "[Pipecat Client] Bot version is less than 1.0.0, which may not be compatible with this client."
```
kompfner, line 604: So the client is only going to check whether the server (the bot) is too old, not whether it's too new for the client? (This is a future-facing concern, and a question about what our general versioning approach will be)

client-js/client/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (window?.navigator?.userAgent) {
+    try {
+      navAgentInfo = Bowser.parse(window.navigator.userAgent);
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (_) {
+      // void
+    }
+  }
+
+  if (navAgentInfo?.browser?.name) {
```
kompfner, line 42: Not reviewing this part too closely, but I'm assuming it's pretty much the same as what we do in daily-js's logging and telemetry code, which has been thoroughly battle-tested?

client-js/rtvi/events.ts
kompfner: I've definitely seen `/** ... */` denote larger sections. But they're usually spread across multiple lines to denote their bigness, otherwise there's no much visual distinction.

```
/**
 * SECTION
 */

// comment 1
// comment 2

/** 
 * NEXT SECTION
 */
```

client-js/rtvi/messages.ts
```diff
@@ -0,0 +1,226 @@
...
+  version: string;
+  about?: unknown; // Optional about data from the bot
+};
+
+type PlatformDetailsValue = undefined | string | number | boolean;
+type NestedPlatformDetails =
+  | PlatformDetailsValue
+  | Record<string, PlatformDetailsValue>;
+
+// This is an interface so that different client libraries can provide their own
```
kompfner, line 80: by "different client libraries" you mean only this one (the web one) and the react native one, which is based on this one?

kompfner (approved):

client-js/rtvi/messages.ts
mattieruth: js, react, and react native

client-js/client/utils.ts
mattieruth: so yes, battle-tested

client-js/client/rest_helpers.ts
Regaddi: `object` doesn't allow primitives, like numbers, strings or booleans, which are generally JSON-serializable.
I think it's a good idea to inform about the intent of `requestData` ("should be JSON-serializable") and provide a matching type:

```ts
type Serializable =
  | string
  | number
  | boolean
  | null
  | Serializable[]
  | { [key: number | string]: Serializable };
```

Unless we _require_ `requestData` to be an object, then we could set it to:

```ts
type SerializableRequestData = { [key: number | string]: Serializable };
```

client-js/rtvi/events.ts
Regaddi: Ah, I didn't think of the "small" comments as subsections when reading it first. It makes sense now ☺️

client-js/client/rest_helpers.ts
mattieruth: thanks!

PR #125, opened 2025-07-04
React: consolidate cam, mic & transport state

### LINKED ISSUE #121: React: Changing micEnabled state does not work
Hi, I'm playing around with the react library and the library design looks great so far! I'm trying to use the `@pipecat-ai/client-react` library, but it seems that the state in the application is not correctly managed.

For example, we have the following custom component:

```js
import { useRTVIClientMicControl } from "@pipecat-ai/client-react";

export function MuteButton() {
  const { enableMic, isMicEnabled } = useRTVIClientMicControl();

  console.log("is mic enabled:", isMicEnabled);

  return (
    <button onClick={() => enableMic(!isMicEnabled)}>
      {isMicEnabled ? "Disable Microphone" : "Enable Microphone"}
    </button>
  );
}
```

I would assume that if I have two of these buttons and click on one of them, that the text of the other button also changes. However that is not the case:

Image

Also, the first click on a button usually enables the microphone, but a second click won't disable the microphone, even though the button text changes.
Interestingly, it will trigger "track started" events, though.

Do you have an idea how to solve these issues?

Regaddi:
Instead of maintaining cam, mic & transport state individually per instance these are now maintained in a surrounding `PipecatClientStateProvider` component and shared via React context.

Fixes #121.

### DIFF
--- a/client-react/src/PipecatClientState.tsx
+++ b/client-react/src/PipecatClientState.tsx
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2024, Daily.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { RTVIEvent, TransportState } from "@pipecat-ai/client-js";
+import React, { createContext, useCallback, useState } from "react";
+
+import { usePipecatClient } from "./usePipecatClient";
+import { useRTVIClientEvent } from "./useRTVIClientEvent";
+
+export const PipecatClientCamStateContext = createContext<{
+  enableCam: (enabled: boolean) => void;
+  isCamEnabled: boolean;
+}>({
+  enableCam: () => {
+    throw new Error(
+      "RTVICamStateContext: enableCam() called outside of provider"
+    );
+  },
+  isCamEnabled: false,
+});
+export const PipecatClientMicStateContext = createContext<{
+  enableMic: (enabled: boolean) => void;
+  isMicEnabled: boolean;
+}>({
+  enableMic: () => {
+    throw new Error(
+      "RTVIMicStateContext: enableMic() called outside of provider"
+    );
+  },
+  isMicEnabled: false,
+});
+export const PipecatClientTransportStateContext =
+  createContext<TransportState>("disconnected");
+
+export const PipecatClientStateProvider: React.FC<React.PropsWithChildren> = ({
+  children,
+}) => {
+  const client = usePipecatClient();
+  const [isCamEnabled, setIsCamEnabled] = useState(false);
+  const [isMicEnabled, setIsMicEnabled] = useState(false);
+  const [transportState, setTransportState] =
+    useState<TransportState>("disconnected");
+
+  useRTVIClientEvent(RTVIEvent.TransportStateChanged, (state) => {
+    setTransportState(state);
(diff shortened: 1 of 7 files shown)


client-react/src/PipecatClientState.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { usePipecatClient } from "./usePipecatClient";
+import { useRTVIClientEvent } from "./useRTVIClientEvent";
+
+export const PipecatClientCamStateContext = createContext<{
+  enableCam: (enabled: boolean) => void;
+  isCamEnabled: boolean;
+}>({
+  enableCam: () => {
+    throw new Error(
+      "RTVICamStateContext: enableCam() called outside of provider"
```
mattieruth, line 19:
```suggestion
      "PipecatClientCamStateContext: enableCam() called outside of provider"
```

client-react/src/PipecatClientState.tsx
```diff
@@ -0,0 +1,84 @@
...
+  },
+  isCamEnabled: false,
+});
+export const PipecatClientMicStateContext = createContext<{
+  enableMic: (enabled: boolean) => void;
+  isMicEnabled: boolean;
+}>({
+  enableMic: () => {
+    throw new Error(
+      "RTVIMicStateContext: enableMic() called outside of provider"
```
mattieruth, line 30:
```suggestion
      "PipecatClientMicStateContext: enableMic() called outside of provider"
```

client-react/src/usePipecatClientCamControl.ts
```diff
@@ -1,41 +1,14 @@
...
-    },
-    [client]
-  );
-
-  return {
-    enableCam,
-    isCamEnabled,
-  };
-};
+export const useRTVIClientCamControl = () =>
```
mattieruth, line 13:
```suggestion
export const usePipecatClientCamControl = () =>
```

client-react/src/usePipecatClientMicControl.ts
```diff
@@ -1,41 +1,15 @@
...
-
-  return {
-    enableMic,
-    isMicEnabled,
-  };
-};
+/**
+ * Hook to control microphone state
+ */
+export const useRTVIClientMicControl = () =>
```
mattieruth, line 14:
```suggestion
export const usePipecatClientMicControl = () =>
```

client-react/src/usePipecatClientTransportState.ts
```diff
@@ -4,17 +4,9 @@
...
-const transportStateAtom = atom<TransportState>("disconnected");
-
-export const usePipecatClientTransportState = () => {
-  const [transportState, setTransportState] = useAtom(transportStateAtom);
-
-  useRTVIClientEvent(RTVIEvent.TransportStateChanged, setTransportState);
-
-  return transportState;
-};
+export const useRTVIClientTransportState = () =>
```
mattieruth, line 11:
```suggestion
export const usePipecatClientTransportState = () =>
```

mattieruth (approved):
i think this looks good. just naming updates needed

Regaddi:
> just naming updates needed

I must have overlooked these when resolving the merge conflicts, thanks for catching!

PR #134, opened 2025-08-01
Add support for new deviceError callback and event

mattieruth:
Introducing a new PipecatClient callback for receiving device errors (mic not found, permissions lacking, etc.).


filipi87 (approved):
LGTM!

PR #139, opened 2025-08-06
Improved usability around connect()

mattieruth:
Deprecated use of connect() for both starting the bot and connecting, separating it out into startBot() and connect() along with a convenience method for doing both: startBotAndConnect()

Also renamed ConnectionEndpoint to a more generic and informative name of APIEndpoint

### DIFF
--- a/client-js/client/client.ts
+++ b/client-js/client/client.ts
@@ -34,10 +34,10 @@ import { transportReady } from "./decorators";
 import { MessageDispatcher } from "./dispatcher";
 import { logger, LogLevel } from "./logger";
 import {
-  APIEndpoint,
+  APIRequest,
   ConnectionEndpoint,
-  getTransportConnectionParams,
-  isAPIEndpoint,
+  isAPIRequest,
+  makeRequest,
 } from "./rest_helpers";
 import {
   Tracks,
@@ -149,7 +149,7 @@ abstract class RTVIEventEmitter extends (EventEmitter as unknown as new () => Ty
 
 export class PipecatClient extends RTVIEventEmitter {
   protected _options: PipecatClientOptions;
-  private _startResolve: ((value: unknown) => void) | undefined;
+  private _connectResolve: ((value: BotReadyData) => void) | undefined;
   protected _transport: Transport;
   protected _transportWrapper: TransportWrapper;
   protected declare _messageDispatcher: MessageDispatcher;
@@ -360,9 +360,7 @@ export class PipecatClient extends RTVIEventEmitter {
    * @param startBotParams
    * @returns Promise that resolves to TransportConnectionParams or unknown
    */
-  public async startBot(
-    startBotParams: APIEndpoint
-  ): Promise<TransportConnectionParams | unknown> {
+  public async startBot(startBotParams: APIRequest): Promise<unknown> {
     if (
       ["authenticating", "connecting", "connected", "ready"].includes(
         this._transport.state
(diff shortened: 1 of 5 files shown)


client-js/client/rest_helpers.ts
```diff
@@ -10,26 +10,36 @@ type Serializable =
   | Serializable[]
   | { [key: number | string]: Serializable };
 
-export interface ConnectionEndpoint {
-  endpoint: string;
+export interface APIEndpoint {
```
kompfner, line 13: maybe `APIEndpointParams`? it specifies more than just the endpoint itself (as evidenced by the argument name `startBotParams`)

client-js/client/client.ts
```diff
@@ -353,15 +354,44 @@ export class PipecatClient extends RTVIEventEmitter {
     await this._transport.initDevices();
   }
 
+  /**
+   * startBot() is a method that initiates the bot by posting to a specified endpoint
+   * that optionally returns connection parameters for establishing a transport session.
+   * @param startBotParams
+   * @returns Promise that resolves to TransportConnectionParams or unknown
```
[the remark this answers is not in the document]
kompfner: wait...looks like `getTransportConnectionParams` contractually returns a `TransportConnectionParams`. so i'm not following where the `unknown` comes in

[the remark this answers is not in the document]
kompfner: (maybe `getTransportConnectionParams` needs to update _its_ contract?)

client-js/client/client.ts
```diff
@@ -353,15 +354,44 @@ export class PipecatClient extends RTVIEventEmitter {
...
+        this._transport.state
+      )
+    ) {
+      throw new RTVIErrors.RTVIError(
+        "Voice client has already been started. Please call disconnect() before starting again."
+      );
+    }
+    this._transport.state = "authenticating";
+    this._abortController = new AbortController();
+    const response = await getTransportConnectionParams(
```
kompfner, line 377: now that we're being more opinionated/prescriptive about the fact that the endpoint that gets the connection params _is_ the "start bot" endpoint, updating this method name accordingly would read more clearly, imo

kompfner: something along the lines of `callStartBotEndpoint` (not wedded to that name, we just what first popped into my head)

client-js/client/client.ts
```diff
@@ -377,6 +407,13 @@ export class PipecatClient extends RTVIEventEmitter {
       );
     }
 
+    if (connectParams && isAPIEndpoint(connectParams)) {
+      logger.warn(
+        "Calling connect with an API endpoint is deprecated. Use startBotAndConnect() instead."
```
kompfner, line 412: 👍

client-js/client/rest_helpers.ts
kompfner: i think "params" in the name does add information—it clarifies that it's not just a url or path, it contains params for making the request

client-js/client/client.ts
mattieruth: I was actually thinking of going the other way with this and renaming it to something like `makeRequest`

client-js/client/client.ts
```diff
@@ -413,6 +437,13 @@ export class PipecatClient extends RTVIEventEmitter {
     });
   }
 
+  public async startBotAndConnect(
+    startBotParams: APIEndpoint
+  ): Promise<unknown> {
```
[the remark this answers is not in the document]
kompfner: oh wait. no, doesn't it return what `_startResolve` resolves to? `BotReadyData`?

client-js/client/client.ts
```diff
@@ -353,15 +354,44 @@ export class PipecatClient extends RTVIEventEmitter {
...
 
+  /**
+   * startBot() is a method that initiates the bot by posting to a specified endpoint
+   * that optionally returns connection parameters for establishing a transport session.
+   * @param startBotParams
+   * @returns Promise that resolves to TransportConnectionParams or unknown
+   */
+  public async startBot(
+    startBotParams: APIEndpoint
+  ): Promise<TransportConnectionParams | unknown> {
```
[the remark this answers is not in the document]
kompfner: i'm wrong! i was confused, thinking about how you don't have to explicitly _construct_ a `Promise` within an `async` method. ignore me!

client-js/client/rest_helpers.ts
mattieruth: going with `APIRequest`

client-js/client/client.ts
mattieruth: updating that as well to be generic and moving the `StartBotError` creation into `startBot()` itself...

kompfner (approved):
🥳

PR #141, opened 2025-08-13
Add the check for initialized devices to startBotAndConnect to ensure…

mattieruth:
… the default behavior of turning on the devices when connecting works


filipi87 (approved):
LGTM!

PR #152, opened 2025-09-30
Add an optional field, source, to the BotTranscript message

mattieruth:


marcus-daily:
@mattieruth Is the difference between the two constants that: if the source is "tts" then it was spoken by the bot, whereas if it was "llm" then it wasn't spoken? It's a bit confusing because the LLM is ultimately the source of everything the bot says.

Would it be better to have e.g. a boolean flag saying whether the text was spoken?

mattieruth:
slowly backing away from this change. marcus' comment led to a 2hour discussion without a clear answer for how bot transcriptions and output should work, so the types are TBD :)

Issue #114, opened 2025-05-11
SSR Error When Compiling with Server Side Vite


mwood23:
I'm getting this error when trying to consume `@pipecat-ai/client-js` 

```
Error in renderToPipeableStream: SyntaxError: [vite] Named export 'LogLevel' not found. The requested module '@pipecat-ai/client-js' is a CommonJS module, which may not
 support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@pipecat-ai/client-js';
const {RTVIClient, LLMHelper, LogLevel} = pkg;
```

I'm not sure if CJS is preferred for some reason or not, but using `"type": "module"` in the package.json would fix it. If CJS is needed for some reason, can export maps be added? This fixes it for me:

```patch
diff --git a/package.json b/package.json
index 75195d82731f2baae100c293c071f5e0937d215d..3131e614de8f635259e7f50554e6c7f2b2063632 100644
--- a/package.json
+++ b/package.json
@@ -10,6 +10,10 @@
     "type": "git",
     "url": "git+https://github.com/pipecat-ai/pipecat-client-web.git"
   },
+  "exports": {
+    "import": "./dist/index.module.js",
+    "require": "./dist/index.js"
+  },
   "files": [
     "dist",
     "package.json",

```

Note: you will also need to do this for `client-react`, `daily-transport`, and any other module published:

```patch
diff --git a/package.json b/package.json
index 62b64cabce8c8b7e6a7dc463aa50881ae8df65cc..2bdb10e99eefbe7a9661de0744621898a75b91f5 100644
--- a/package.json
+++ b/package.json
@@ -10,6 +10,10 @@
     "type": "git",
     "url": "git+https://github.com/pipecat-ai/pipecat-client-web.git"
   },
+  "exports": {
+    "import": "./dist/index.module.js",
+    "require": "./dist/index.js"
+  },
   "files": [
     "dist",
     "package.json",
```

```patch
diff --git a/package.json b/package.json
index a5ffc1e5be7a839277d66b80ac317671705d73fe..6c471052d887fc8d9374c1cc51fbcc23773c369f 100644
--- a/package.json
+++ b/package.json
@@ -10,6 +10,10 @@
     "type": "git",
     "url": "git+https://github.com/pipecat-ai/pipecat-client-web-transports.git"
   },
+  "exports": {
+    "import": "./dist/index.module.js",
+    "require": "./dist/index.js"
+  },
   "files": [
[...]

PR #166, opened 2025-12-19
Transition to official module support

mattieruth:
@jptaylor @Regaddi -- Per https://github.com/pipecat-ai/pipecat-client-web/issues/114 -- Is there any reason NOT to do this?

### DIFF
--- a/client-js/package.json
+++ b/client-js/package.json
@@ -2,8 +2,8 @@
   "name": "@pipecat-ai/client-js",
   "version": "1.5.0",
   "license": "BSD-2-Clause",
+  "type": "module",
   "main": "dist/index.js",
-  "module": "dist/index.module.js",
   "types": "dist/index.d.ts",
   "source": "index.ts",
   "repository": {


mattieruth:
blech. reading up more on this. The format we've gone with seems to be the preferred way by [parcel](https://parceljs.org/getting-started/library/#commonjs-and-es-modules) -- it mentions that you can use `"type": "module"` but as like an afterthought without saying if anything else needs to be changed or added.

Regaddi:
I wonder if only adding the `exports` field would suffice, as proposed in #114.

Regaddi:
@mattieruth Claude seems to err on the side of adding `exports`, too:

---

## The Problem (Issue #114)
When Vite does SSR, it resolves @pipecat-ai/client-js to the CJS bundle (dist/index.js via "main"), because there's no "exports" map to tell Node/Vite to pick the ESM file. CJS named exports can't be destructured in an ESM context, hence:

```
Named export 'LogLevel' not found. The requested module '@pipecat-ai/client-js' is a CommonJS module
```
## Potential Problems with "type": "module"

1. **Breaks the current CJS output.** Right now Parcel generates [dist/index.js](vscode-webview://0ppua1g7evhtk01lsei8bc9ec6b5hfpg1u2qaniv2knsc26h96p9/client-js/dist/index.js) with `require()` / `module.exports` syntax. With `"type": "module"`, Node treats all `.js` files as ESM — so Node would choke on the CJS syntax in that file. Parcel would need to either stop generating CJS, or output it as `.cjs` instead.

2. **Drops CJS consumers.** Anyone doing `const { RTVIClient } = require('@pipecat-ai/client-js')` breaks if the CJS build goes away. This is less common in 2026 but still possible in older Node codebases, test runners, or tooling.

3. **Jest may break.** The current test setup uses `ts-jest` + `jsdom`. Jest's ESM support has historically been experimental/fragile. Switching to `"type": "module"` often requires `--experimental-vm-modules`, changes to `ts-jest` config, or migrating to a different test runner.

4. **Coordinated multi-package change.** The issue notes this must be done across `client-js`, `client-react`, `daily-transport`, and other transport packages simultaneously, or you get mismatched module formats between interdependent packages.

5. **Parcel's own behavior changes.** PR #166 removes the `"module"` field and keeps only `"main"`. Parcel's library mode uses these fields to decide what to output. The build output may change in unexpected ways — needs testing.

## The Simpler Fix: exports Map

Both the issue reporter and @Regaddi in [PR #166's comments](https://github.com/pipecat-ai/pipecat-client-web/pull/166) suggest the same thing — just add an `exports` field **without** `"type": "module"`:

```
"exports": {
  ".": {
    "types": "./dist/index.d.ts",
    "import": "./dist/index.module.js",
    "require": "./dist/index.js"
  }
}
```
This tells Node/Vite/bundlers to use the ESM build for `import` and the CJS build for `require`, keeping both formats working. No build changes, no Jest breakage, no CJS consumer breakage. It solves the SSR issue directly.

## Recommendation

The `exports` map is the safer path — it fixes the SSR problem without any of the downsides above. If you want to eventually go full ESM-only (drop CJS), that's a separate, intentional breaking change that should be a major version bump.

PR #167, opened 2026-01-12
New Feature: Support sending files

mattieruth:
1. adds new RTVI types
2. adds new `sendFile()` method
3. adds better error handling for sending messages that are too large
4. Currently, only sending files as bytes is tested and working

### DIFF
--- a/client-js/client/client.ts
+++ b/client-js/client/client.ts
@@ -29,13 +29,17 @@ import {
   LLMFunctionCallStartedData,
   LLMFunctionCallStoppedData,
   MediaState,
+  MimeTypeMapping,
   Participant,
   PipecatMetricsData,
   RTVI_PROTOCOL_VERSION,
   RTVIEvent,
   RTVIEvents,
+  RTVIFile,
+  RTVIFileFormat,
   RTVIMessage,
   RTVIMessageType,
+  SendFileOptions,
   SendTextOptions,
   setAboutClient,
   TranscriptData,
@@ -1091,7 +1095,9 @@ export class PipecatClient extends RTVIEventEmitter {
   public cancelUIJobGroup(jobId: string, reason?: string): void {
     const payload: UICancelJobGroupData = { job_id: jobId };
     if (reason !== undefined) payload.reason = reason;
-    this._sendMessage(new RTVIMessage(RTVIMessageType.UI_CANCEL_JOB_GROUP, payload));
+    this._sendMessage(
+      new RTVIMessage(RTVIMessageType.UI_CANCEL_JOB_GROUP, payload)
+    );
   }
 
   /**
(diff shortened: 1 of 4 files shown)


client-js/rtvi/messages.ts
```diff
@@ -262,6 +263,117 @@ export type SendTextOptions = {
...
+};
+
+export type FileSourceType = "bytes" | "url" | "id";
+
+export type FileBytes = {
+  type: Extract<FileSourceType, "bytes">;
+  bytes: string;
+  width?: number;
+  height?: number;
+};
```
filipi87, line 347: It does feels a bit odd having width and height inside `FileBytes`. I believe this is needed in case we are sending images, right ? 

Maybe do something like this:
```
type ImageFileBytes = FileBytes & {
  width: number;
  height: number;
};
```

or

```
type FileBytes = {
  type: "bytes";
  bytes: string;
  metadata?: ImageMetadata;
};

type ImageMetadata = {
  width: number;
  height: number;
};
```

What do you think ?

client-js/rtvi/messages.ts
```diff
@@ -262,6 +263,117 @@ export type SendTextOptions = {
...
+  // RTVI definition takes the Mime type here, but in client-js, we support
+  // clients providing shorthands defined above and we map them to Mime types
+  format: string;
+  source: FileBytes | FileUrl | FileId;
+};
+
+export type SendFileOptions = {
+  run_immediately?: boolean;
+  audio_response?: boolean;
+  // for things like 'detail' in openAI or 'citations' in Bedrock
```
filipi87, line 369: 👍

client-js/rtvi/messages.ts
```diff
@@ -262,6 +263,117 @@ export type SendTextOptions = {
...
+  custom_options?: { [key: number | string]: Serializable };
+};
+
+export type FileSupport = {
+  formats: string[];
+  sources: FileSourceType[];
+  maxSize: number; // bytes
+};
```
filipi87, line 377: It looks like we are not using this anywhere. Should we keep it?

client-js/rtvi/messages.ts
```diff
@@ -262,6 +263,117 @@ export type SendTextOptions = {
...
+  type: Extract<FileSourceType, "id">;
+  id: string;
+};
+
+export type RTVIFile = {
+  name?: string;
+  // RTVI definition takes the Mime type here, but in client-js, we support
+  // clients providing shorthands defined above and we map them to Mime types
+  format: string;
+  source: FileBytes | FileUrl | FileId;
```
filipi87, line 363: In summary, we're basically sending either the file itself (as bytes), a URL to a file, or the ID of a file that was previously uploaded somewhere. Is that correct ?

I don't see `FileId` being used anywhere else. Are we planning to support it directly ?

client-js/client/client.ts
```diff
@@ -1155,6 +1161,187 @@ export class PipecatClient extends RTVIEventEmitter {
...
 
+  @transportReady
+  public async sendFile(
+    file: RTVIFile | File,
+    content: string,
+    options: SendFileOptions = {}
+  ) {
+    if (
+      this._botVersion[0] < 1 ||
+      (this._botVersion[0] === 1 && this._botVersion[1] < 3)
```
filipi87: Shouldn't this check be like this to use this feature ?
```suggestion
      this._botVersion[0] < 2 || (this._botVersion[0] === 2 && this._botVersion[1] < 2)
```

client-js/client/client.ts
```diff
@@ -1155,6 +1161,187 @@ export class PipecatClient extends RTVIEventEmitter {
...
+    if (file instanceof File) {
+      // Estimate the message size with base64 encoding overhead (~33% larger)
+      // Add buffer for message wrapper overhead. This saves us from having to
+      // unnecessarily read the file into memory and encode it to base64.
+      const estimatedEncodedSize = Math.ceil(file.size * 1.37) + 1000;
+
+      if (estimatedEncodedSize > this._transport.maxMessageSize) {
+        uploadFile = file;
+      } else {
+        return new Promise<void>((resolve) => {
```
filipi87: Suggestion from Claude:

  _The inline-file-read path returns `new Promise<void>((resolve) => { ... })` with an `async reader.onload` handler, and never rejects. Three error paths leave the promise permanently unsettled, so `await client.sendFile(...)` hangs forever:

  - the bare `throw new RTVIErrors.RTVIError("Could not read file data")` when `e.target?.result` is missing (`client.ts:1211`) rejects only the handler's own internal promise, which `FileReader` never awaits;
  - `await sendFileMessage()` (`client.ts:1223`) can reject — `_sendMessage` throws `MessageTooLargeError` when `messageSizeWithinLimit` fails (`client.ts:806`), which is reachable because the `file.size * 1.37` branch decision is only a heuristic;
  - no `reader.onerror` is registered, so any `FileReader` failure never settles the promise either.

  Suggested fix: wrap the `onload` body in try/catch and call `reject(err)` on failure; add `reader.onerror = (err) => reject(err)`._

client-js/rtvi/messages.ts
```diff
@@ -262,6 +263,117 @@ export type SendTextOptions = {
...
+  | { [key: number | string]: Serializable };
+
+export type RTVIImageFormat =
+  | "png"
+  | "jpg"
+  | "jpeg"
+  | "webp"
+  | "gif"
+  | "heic"
+  | "hief";
```
filipi87: Typo: I believe it should be `heif`.

client-js/rtvi/messages.ts
```diff
@@ -262,6 +263,117 @@ export type SendTextOptions = {
...
+
+export const MimeTypeMapping: Record<RTVIFileFormat, string> = {
+  // Images
+  png: "image/png",
+  jpg: "image/jpeg",
+  jpeg: "image/jpeg",
+  webp: "image/webp",
+  gif: "image/gif",
+  heic: "image/heic",
+  hief: "image/heif",
```
filipi87: Same here: `heif`

client-js/client/client.ts
```diff
@@ -1155,6 +1161,187 @@ export class PipecatClient extends RTVIEventEmitter {
     );
   }
 
+  @transportReady
+  public async sendFile(
```
filipi87, line 1165: This function is a bit hard to follow. It is doing three things at once: normalizing the input, deciding inline vs. upload, and sending.

Would it be worth restructuring it ? Maybe something like:

```
  public async sendFile(file: RTVIFile | File, content: string, options: SendFileOptions = {}) {
    this.assertBotSupportsSendFile();

    const resolved = file instanceof File
      ? await this.resolveBrowserFile(file)
      : await this.resolveRTVIFile(file);

    await this._sendMessage(
      new RTVIMessage(RTVIMessageType.SEND_FILE, { file: resolved, content, options })
    );
  }

  /** Small files go inline as base64; anything larger is uploaded first. */
  private async resolveBrowserFile(file: File): Promise<RTVIFile> {
    if (this.exceedsMessageLimit(estimateBase64Size(file.size))) {
      return this.uploadFile(file);
    }
    return {
      name: file.name,
      format: file.type,
      source: { type: "bytes", bytes: await readAsBase64(file) },
    };
  }

  /** Only inline byte payloads need resolving — url and id sources are already remote. */
  private async resolveRTVIFile(file: RTVIFile): Promise<RTVIFile> {
    const normalized = { ...file, format: toMimeType(file.format) };
    if (normalized.source.type !== "bytes") return normalized;
    if (!this.exceedsMessageLimit(normalized.source.bytes.length)) return normalized;
    return this.uploadFile(toUploadableFile(normalized));
  }

```

mattieruth: i like it.

client-js/rtvi/messages.ts
mattieruth: i like it, though, even for images width/height is optional. TBH, maybe they can just go away? I originally had it because I thought it was required in some scenarios. In pipecat, `size` [is required as part of the `UserImageRawFrame`](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/frames/frames.py#L177) that gets generated from the message, but as an [already encoded image, the value isn't used](https://github.com/pipecat-ai/pipecat/blob/main/src/pipecat/processors/aggregators/llm_context.py#L149).

In the RTVI processor, I just [default the size to 0x0](https://github.com/pipecat-ai/pipecat/blob/150ced4921ae588517190084939112e85ace7965/src/pipecat/processors/frameworks/rtvi/processor.py#L587-L590) if it's not provided. Maybe we remove this as part of the spec and always use 0x0?

client-js/rtvi/messages.ts
mattieruth: Correct. Right now I think `FileId` only works with a runner that supports uploading larger files and accessing them from on disc. However, I did think that it should be spec'd and written in a way that could support the various llm's File APIs. All of which either take an id (anthropic & openai) or are built in a way i'm not sure makes sense using from a client anyway (gemini).

mattieruth: But for now, [i'm explicitly not supporting it](https://github.com/pipecat-ai/pipecat/blob/150ced4921ae588517190084939112e85ace7965/src/pipecat/processors/frameworks/rtvi/processor.py#L554-L557).  The id must start with `pipecat:`. Now, whether this is the right prefix or approach to this... not sure. I'd love feedback.

client-js/rtvi/messages.ts
mattieruth: this is a TODO to actually support. Right now, these options fall on the floor.

client-js/rtvi/messages.ts
mattieruth: In my original spec, I wanted a way for clients to get support requirements from the server. I haven't built it out yet. So I dunno. Should I? Should I comment it and leave as a todo? The general idea was so that UIs could pre-filter out files that won't work and avoid unnecessary hefty uploads.

client-js/rtvi/messages.ts
filipi87: I think it would be nice to provide a way to send the size to keep consistency with the server side API, even if this isn't currently needed for Pipecat.

client-js/client/client.ts
mattieruth: done!

client-js/rtvi/messages.ts
mattieruth: done!

client-js/rtvi/messages.ts
filipi87: > Should I comment it and leave as a todo? 

I think so. At least it would make it clear that we are not supporting it yet.

client-js/rtvi/messages.ts
filipi87: Agreed. Let's keep it like this for now.

We should probably just document it or create an example in `pipecat-examples` showing how to create this runner.

filipi87 (approved):
LGTM!

Great to see this API out! 🚀

client-js/client/client.ts
mattieruth: done.

client-js/client/client.ts
mattieruth: i believe this was resolved when I did the refactor.

client-js/rtvi/messages.ts
mattieruth: done

client-js/rtvi/messages.ts
mattieruth: done

mattieruth:
Note: I don't think I should merge this until the Pipecat changes are in and released.

PR #175, opened 2026-02-07
Feature: Added support for new RTVI function call events

mattieruth:
RTVI 1.2.0 introduces new function call lifecycle events (started, in-progress, and stopped) with varying degrees of metadata. This commit introduces client-side support for these new methods along deprecation of the llm-function-call event that is superceded by these new ones.


markbackman (approved):
LGTM! Co-pilot found a few things that I would have. Nothing additional to add. Approving, pending the co-pilot feedback.

PR #179, opened 2026-02-25
feat(client-react): built-in conversation state management

Regaddi:
This PR ports the core conversation implementation from https://github.com/pipecat-ai/voice-ui-kit to this repository, so people using the Pipecat React library directly can benefit from a streamlined conversation state, without having to opt into the Voice UI kit.

## What's included

**Zero-config conversation tracking** — `PipecatClientProvider` now wraps children in a `PipecatConversationProvider` automatically. No additional setup or store wiring required.

**Two new hooks:**
- `usePipecatConversation()` — returns the processed message array with karaoke-style `BotOutputText` (spoken/unspoken split) and an `onMessageAdded` callback
- `useConversationContext()` — provides `injectMessage` for programmatic message insertion and `botOutputSupported` for feature detection

**Full BotOutput support** — handles the dual-stream nature of bot responses (unspoken text arrives ahead of TTS, spoken text advances a cursor through it) with fuzzy word matching, contraction handling, mismatch recovery, and premature-finalization un-winding when TTS pauses mid-response.

**Function call lifecycle** — tracks Started → InProgress → Completed with robust out-of-order tolerance, deduplication by `tool_call_id`, and cancellation support.

**Exported utilities** for consumers who want to post-process messages: `sortByCreatedAt`, `filterEmptyMessages`, `mergeMessages`, `deduplicateFunctionCalls`, `isMessageEmpty`.

## Architecture

**RTVI events**
```
→ useConversationEventWiring (timers, space-prepending, finalization)
  → conversationActions (state mutations, normalization pipeline)
    → Jotai atoms (raw state)
      → usePipecatConversation (spoken/unspoken transform)
        → UI
```

**Key design decisions:**
- **Parts-based message model** — each message contains an array of `ConversationMessagePart`, supporting mixed content types (text, code blocks, function calls) with per-part finalization
- **Write-time normalization** — messages are sorted, filtered, deduplicated, and merged on every write, so reads are always clean
- **Premature finalization recovery** — if BotStoppedSpeaking fires mid-response (TTS pause), the message gets finalized early; when more output arrives, the system un-finalizes instead of creating a duplicate bubble

### DIFF
--- a/client-react/src/conversation/botOutput.ts
+++ b/client-react/src/conversation/botOutput.ts
@@ -72,7 +72,10 @@ const findSpokenPositionInUnspoken = (
     .filter(Boolean);
 
   // Sequential match, allowing prefix match for contractions (e.g. "I" vs "I'm")
+  // and limited skipping of mismatched unspoken words (e.g. punctuation artifacts).
   let matchedWords = 0;
+  let consecutiveSkips = 0;
+  const MAX_CONSECUTIVE_SKIPS = 2;
   for (
     let i = 0;
     i < unspokenWords.length && matchedWords < spokenWords.length;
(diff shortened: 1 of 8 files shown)


Regaddi:
> since this is porting, i think it all makes sense. I'm really not the React judge. I think my only question is about function calling being part of the Conversation. I guess this is more of a philosophical question but it seems odd in that context. I think of it as more part of just the overall event timeline... but I guess, maybe if you consider it synonymous with "thinking" ?? is that the reason?

So one of the main reasons to include function calls is because they are part of the conversation flow. They happen between specific messages, and that positional context matters for rendering. Keeping them in the same timeline avoids consumers having to reconcile two separate streams to get the correct interleaved view.

The API is designed so they can be filtered out (`noFunctionCalls` prop) or customized (`functionCallRenderer`) at the component level. And the "thinking" analogy is basically right: function calls are what the bot does between speaking turns. Showing them inline lets developers build UIs like "Searching for flights..." or a weather card mid-conversation, which only makes sense in context.

PR #180, opened 2026-03-03
Add conventional commits, Release Please, and OIDC publishing

Regaddi:
I'm suggesting this setup, which we've already established in the [Voice UI Kit](https://github.com/pipecat-ai/voice-ui-kit). It removes most of the manual overhead for managing releases.

---

## Summary

Adds automated release infrastructure to replace the manual bump-and-publish workflow:

- **Conventional commits** enforced via Husky + commitlint with `js` and `react` scopes (scopeless also allowed for root-level changes)
- **lint-staged** runs ESLint on staged `.ts/.tsx` files in both packages on pre-commit
- **Release Please** configured for independent per-package versioning with separate tags (`client-js-v*`, `client-react-v*`) and auto-generated changelogs
- **OIDC trusted publishing** — publish jobs use `--provenance` with `id-token: write` instead of `NPM_TOKEN` secret
- **Publish RC workflow** retained as a manual dispatch for prerelease versions

### How it works

1. Conventional commits on `main` trigger Release Please to open/update a release PR per package
2. Merging a release PR creates a GitHub Release and triggers the publish job for that package
3. Each publish job runs tests, builds, and publishes with npm provenance

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -2,6 +2,7 @@
   "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
   "release-type": "node",
   "include-component-in-tag": true,
+  "separate-pull-requests": true,
   "packages": {
     "client-js": {
       "component": "client-js",


.commitlintrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": ["@commitlint/config-conventional"],
```
mattieruth: can you add to the README what the commit structure needs to be now? what's the affect of this?

Regaddi: I added examples to the CONTRIBUTING file.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,66 @@
+name: Release Please
```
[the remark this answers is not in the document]
Regaddi: Yay, there's an option to split into separate PRs per package 🙌 
I just enabled that here and in the transports package, because I think it makes sense to have fine-grained release control for each package, e.g. shipping a React patch shouldn't be blocked by a minor change in client-js and vice-versa.

PR #195, opened 2026-03-26
fix(react): backdate mid-turn function calls to preserve karaoke cursor

Regaddi:
## Summary
- When the LLM calls a function mid-turn, the function call message inserts chronologically after the active assistant message, causing subsequent spoken botOutput to create a new message instead of advancing the existing karaoke cursor
- `addFunctionCall` now backdates its timestamp before the active assistant message (same approach as #193 for injected messages), keeping the assistant response as the last message so the cursor continues uninterrupted

## Test plan
- [x] Added tests: backdating during active assistant, no backdating when finalized, no backdating when assistant isn't last
- [x] All 161 tests pass
- [ ] Manual verification with a bot that calls functions mid-response

### DIFF
--- a/client-react/src/conversation/conversationActions.ts
+++ b/client-react/src/conversation/conversationActions.ts
@@ -653,11 +653,28 @@ export function addFunctionCall(
   }
 
   const now = new Date();
+
+  // If the most recent message is an active (non-final) assistant message,
+  // backdate the function call so it sorts before the assistant's createdAt.
+  // This prevents the function call from splitting the bot's response across
+  // two bubbles and breaking the karaoke cursor.
+  const lastMessage = messages[messages.length - 1];
+  let timestamp: string;
+  if (
+    lastMessage?.role === "assistant" &&
+    lastMessage.final === false
+  ) {
+    const assistantTime = new Date(lastMessage.createdAt);
+    timestamp = new Date(assistantTime.getTime() - 1).toISOString();
+  } else {
+    timestamp = now.toISOString();
+  }
+
   const message: ConversationMessage = {
     role: "function_call",
     final: false,
     parts: [],
-    createdAt: now.toISOString(),
+    createdAt: timestamp,
     updatedAt: now.toISOString(),
     functionCall: {
       function_name: data.function_name,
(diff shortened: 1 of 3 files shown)


client-react/src/conversation/conversationActions.ts
```diff
@@ -653,11 +653,28 @@ export function addFunctionCall(
...
+  // This prevents the function call from splitting the bot's response across
+  // two bubbles and breaking the karaoke cursor.
+  const lastMessage = messages[messages.length - 1];
+  let timestamp: string;
+  if (
+    lastMessage?.role === "assistant" &&
+    lastMessage.final === false
+  ) {
+    const assistantTime = new Date(lastMessage.createdAt);
+    timestamp = new Date(assistantTime.getTime() - 1).toISOString();
```
mattieruth, line 668: i guess same question as other PR. is it possible for this to cause the timestamps to be out of order? and if so, what might the fallout of that be?

Regaddi: Let me paste a very confident response 😄 

---

No. The backdating is safe from ordering issues because of two constraints working together:

  1. **We only backdate when the assistant is the last message** — the check is `lastMessage?.role === "assistant" && lastMessage.final === false`, where `lastMessage` is `messages[messages.length - 1]`. Since `normalizeMessagesForUI` sorts by `createdAt`, the assistant has the latest timestamp of all messages.
  2. **The backdated timestamp is assistantTime - 1ms** — this is guaranteed to be after every message that precedes the assistant (they all have earlier timestamps), and before the assistant itself.

So the resulting order is always: `...prior messages... → function call → assistant`, which is correct.

The one nuance is if multiple function calls are backdated during the same assistant turn — they'd all get the same `assistantTime - 1ms` timestamp. But `sortByCreatedAt` is a stable sort (it only reorders when timestamps differ), so they'd maintain insertion order among themselves. All still before the assistant.

mattieruth: i _think_ I follow, but I still have a creeping suspicion that there's a corner case.

mattieruth (approved):

PR #197, opened 2026-04-20
feat(react): add botOutputFilter option to usePipecatConversation

Regaddi:
## Summary

- Adds a `BotOutputFilter` type and a `botOutputFilter` option on `usePipecatConversation` that controls which portions of each `BotOutputText` (`spoken`, `unspoken`) are returned by the hook.
- The underlying atoms still hold the full data; only the hook's output is filtered. Both flags default to `true`, so existing callers are unaffected.

## Motivation

voice-ui-kit has a [new `textRenderMode` switch](https://github.com/pipecat-ai/voice-ui-kit/pull/118) on its `Conversation` / `ConversationPanel` / `ConsoleTemplate` components with three modes:

- `"karaoke"` (default): spoken text normal, unspoken text muted (current behavior).
- `"spoken"`: only show what the bot has already said — this is the mode that needs the hook to drop the `unspoken` portion, otherwise the karaoke rendering shows it all the time.
- `"unspoken"`: the full LLM text, no highlight sweep — doesn't need the filter but works cleanly with it.

The filter lives here instead of in the UI so any consumer of the hook (not just voice-ui-kit) can opt into the behavior.

## Test plan

- [x] `pnpm --filter @pipecat-ai/client-react test` — 162 tests pass, no behavior change in the default path.
- [x] Verified cross-repo: the textRenderMode PR in voice-ui-kit (#118) consumes this option end-to-end.

### DIFF
--- a/client-react/src/conversation/filterBotOutputText.ts
+++ b/client-react/src/conversation/filterBotOutputText.ts
@@ -1,0 +1,22 @@
+/**
+ * Copyright (c) 2024, Daily.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import type { BotOutputFilter, BotOutputText } from "./types";
+
+/**
+ * Apply a {@link BotOutputFilter} to a {@link BotOutputText}, zeroing out any
+ * portion whose flag is explicitly `false`. Both flags default to `true`, so
+ * an omitted filter (or `undefined`) returns the text unchanged.
+ */
+export function filterBotOutputText(
+  text: BotOutputText,
+  filter?: BotOutputFilter
+): BotOutputText {
+  return {
+    spoken: filter?.spoken === false ? "" : text.spoken,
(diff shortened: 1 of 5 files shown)


mattieruth (commented):
Right now this filter is solely used for filtering on spoken/unspoken. Given the name, is it expected that it might filter by aggregation or other ways in the future? If so, then I think this looks good and I'll approve. If not, we might want to consider a different name for all of this

Regaddi:
> Right now this filter is solely used for filtering on spoken/unspoken. Given the name, is it expected that it might filter by aggregation or other ways in the future? If so, then I think this looks good and I'll approve. If not, we might want to consider a different name for all of this

I wanted to keep the naming and API extendable WRT the different things botOutput event payloads can carry, so yes, a future version might include filters on aggregation types or whatever might become part of the event ☺️

mattieruth (approved):
great! Then I approve :)

PR #198, opened 2026-04-20
fix(react): handle S2S races, mid-sentence drift, and multi-word TTS events

Regaddi:
> **Stacked on [#197](https://github.com/pipecat-ai/pipecat-client-web/pull/197)**. Rebase / re-target to `main` once that lands.

## Summary

- Reconciliation between unspoken (LLM aggregator) and spoken (TTS timing) BotOutput events was failing for several timing patterns seen with OpenAI Realtime and other S2S pipelines, producing duplicated text, stalled cursors, and entire sentences rendering as unspoken in the hook's returned messages.
- Adds two complementary mechanisms in `conversationActions.updateAssistantBotOutput`: character-level prefix absorption of spoken-only fallback parts when an unspoken event arrives, and a gated drop for unmatched spoken events based on whether any unspoken content is still pending.
- Rebuilds the position converter in `findSpokenPositionInUnspoken` to stop drifting one word behind when matching required skipping, and to treat whitespace-separated chunks as single words so contractions (`shouldn't`, `I'll`) don't over-count.
- Splits multi-word spoken events (`Hello! I'm`, `a friendly`, `the open-source`) at token boundaries inside the store so a single TTS chunk can advance across unspoken-part boundaries.
- Adds a `BotOutputEvent` ring buffer for debugging/replay (`botOutputEventsAtom` + `usePipecatConversation().botOutputEvents`).

This is a direct port of [voice-ui-kit#117](https://github.com/pipecat-ai/voice-ui-kit/pull/117) into `conversationActions` / `botOutput` / `usePipecatConversation`.

## Scenarios covered

| # | Scenario | Failure before | Behavior after |
|---|---|---|---|
| A | S2S race: spoken `"Hello"/"there"/"!"` arrive before unspoken `"Hello there!"` | Duplicated `"Hello there !Hello there!"` | Fallbacks absorbed into the single sentence part |
| B | TTS sub-word splits (`Pi`/`pec`/`at`) mid-sentence after unspoken already present | Stray `"pec at documentation for details"` tail | Unmatched tokens dropped mid-stream |
| C | Spoken skipping over earlier words (`shouldn't`, `I'll`) | Cursor lagged one word per skip, sentence end rendered unspoken | Cursor lands past the actual matched word |
| D | Spoken `"Give me just a"` arrives before sentence 2's unspoken | Words dropped as noise; UI stuck on sentence 1 | Kept as fallback, absorbed when sentence 2's unspoken lands |
| E | Multi-word `"Hello! I'm"` crossing a sentence boundary | Whole event discarded; subsequent words lost too | Split into two tokens, first advances sentence 1, second becomes fallback that sentence 2's unspoken absorbs |

## Test plan

- [x] `pnpm --filter @pipecat-ai/client-react test` — 173 tests pass (2 new unit tests for cursor drift + contractions, 9 new integration tests for the scenarios above; mirrors the voice-ui-kit test coverage).
- [x] Existing cascaded-pipeline and spoken-only bot tests still pass unchanged — the new logic is inert in those paths.

### DIFF
--- a/client-react/src/conversation/botOutput.ts
+++ b/client-react/src/conversation/botOutput.ts
@@ -14,6 +14,18 @@ export type BotOutputMessageCursor = {
    * to be spoken, or mismatch recovery advanced past it).
    */
   partFinalFlags: boolean[];
+  /**
+   * Per-part: true if the part was created as a spoken-only fallback (a
+   * spoken event arrived with no matching unspoken content). Such parts
+   * can be absorbed when an unspoken event later arrives that spans them.
+   */
+  partSpokenOnly: boolean[];
+  /**
+   * True once any unspoken event has been received for this message. When
+   * true, unmatched spoken events are dropped instead of spawning fallback
+   * parts (since the unspoken stream is authoritative for this message).
+   */
+  hasReceivedUnspoken: boolean;
 };
 
 export const normalizeForMatching = (text: string): string => {
@@ -75,6 +87,7 @@ const findSpokenPositionInUnspoken = (
   // and limited skipping of mismatched unspoken words (e.g. punctuation artifacts).
   let matchedWords = 0;
   let consecutiveSkips = 0;
+  let lastMatchedUnspokenIdx = -1;
   const MAX_CONSECUTIVE_SKIPS = 2;
   for (
     let i = 0;
@@ -86,6 +99,7 @@ const findSpokenPositionInUnspoken = (
     if (candidate === target || candidate.startsWith(target)) {
       matchedWords++;
       consecutiveSkips = 0;
+      lastMatchedUnspokenIdx = i;
       continue;
     }
     consecutiveSkips++;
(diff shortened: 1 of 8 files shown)


mattieruth (commented):
I think I mostly have to take your word for the logic here, but want to make sure we aren't overlooking the case where text is really meant to be unspoken. There is always the possibility that bot output was generated from text with `skip_tts: true`, meaning that the text is not ever expected to have a version `spoken: true`. Then of course, there's always text meant to be spoken but interrupted. Do we have tests to cover those scenarios to make sure these changes don't affect that?

Side note: while reading this, along with the previous filtering PR, i'm wondering if we should consider adding another flag in botOuput that's present whenever `spoken: false` to indicate whether it is being sent to the TTS to be spoken or not. Question is: whether we actually know that. for the most part, i think we do. Also, not sure it really helps since you still have to handle interruptions

Regaddi:
Addressed the review feedback in [8a7cf92](https://github.com/pipecat-ai/pipecat-client-web/pull/198/commits/8a7cf92) with three new integration tests specifically for the scenarios you called out:

**`skip_tts`-style / non-spoken aggregation parts**
- `never absorbs an existing non-spoken aggregation part into a later unspoken event` — an unspoken event with a non-default `aggregatedBy` (e.g. `"code"`) is flagged `partSpokenOnly: false`, so the absorption walk stops at it; a later unspoken sentence creates a new part instead of eating the code block.
- `lets mismatch-recovery skip a non-spoken part when spoken words only match the next sentence` — the pre-existing mismatch-recovery path is still used for the cascaded case where spoken words only match surrounding sentences. Confirms the new drop-guard doesn't fire here, so no stray fallbacks are spawned.

**Interruption**
- `preserves the absorbed cursor offset when the assistant is interrupted mid-utterance` — S2S race absorbs a prefix, user interrupts before the rest is spoken. The cursor offset inside the absorbed part survives finalization (so the UI can still render the partially-spoken prefix of the interrupted message).

Re: the side note about adding a `skip_tts` / "will be spoken" flag to `BotOutput` — I think you're right that in practice it'd have limited payoff once interruptions enter the picture (we still need the best-effort matcher). Today consumers can mark content non-spoken via `aggregationMetadata.isSpoken: false` on a custom `aggregatedBy`, which is what the existing `"code"` test exercises. Happy to follow up with a proper flag if you decide it's worthwhile — it would be a small pipecat-client-web + pipecat server change, plus a one-line cursor guard here (skip the part entirely instead of relying on mismatch-recovery).

Test count: 185 (was 173). `pnpm --filter @pipecat-ai/client-react test`.

mattieruth (approved):
I have the feeling we'll continue to find corner case issues but i'm excited for this.

Regaddi:
> I have the feeling we'll continue to find corner case issues but i'm excited for this.

Me too 😄 The good thing is that whenever we run into this, we can add tests to the test suite. I'm constantly adding to this, so this is only getting better and better 🤩

PR #202, opened 2026-04-24
test(js): characterize initDevices() and device-error behavior

Regaddi:
## Summary

Plan A, step 1 (part 1 of 2). Adds characterization tests that lock in today's `PipecatClient` device-init behavior before the `MediaState` refactor lands in step 2. Companion PR in `pipecat-client-web-transports` will cover the per-transport matrix (Daily, SmallWebRTC, OpenAI, WebSocket, Gemini Live).

New suite: `client-js/tests/initDevices.spec.ts`. Covers:

- Initial `TransportState` is `"disconnected"` (same sentinel as post-disconnect — the root ambiguity Plan A targets).
- Explicit `initDevices()`: `disconnected` → `initializing` → `initialized`.
- Repeated explicit `initDevices()` calls re-enter the `initializing` state.
- Implicit init via `connect()` and `startBot()` when state is `"disconnected"`.
- The implicit-init gate is skipped when state is already `"initialized"`.
- Reconnect after disconnect re-runs `initDevices()` (state reverts to `"disconnected"`).
- `initDevices()` rejection leaves state at `"initializing"` (stuck state visible to consumers).
- `RTVIEvent.DeviceError` propagation for every `DeviceErrorType` (`in-use`, `permissions`, `undefined-mediadevices`, `not-found`, `constraints`, `unknown`) and every `devices` combination.
- `DeviceError` does not mutate `TransportState`.

Also documents a latent bug in a comment: `connect()` hangs when `initDevices()` rejects because the `await` sits outside the `try/catch` in `client.ts`. Not asserted as a passing test (triggering the path produces an unhandled rejection). Step 2 will handle the error deliberately and re-characterize.

No production code changes. No version bump.

## Test plan

- [x] `npx jest` — all 78 tests pass (22 new)
- [x] `npx tsc --noEmit` — clean
- [x] `npx eslint client/ rtvi/ tests/` — clean (pre-existing errors in committed `dist/` are unrelated)

### DIFF
--- a/client-js/tests/initDevices.spec.ts
+++ b/client-js/tests/initDevices.spec.ts
@@ -1,0 +1,321 @@
+/**
+ * Copyright (c) 2024, Daily.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Characterization tests for initDevices() and the device-state surface on
+ * PipecatClient. These lock in today's behavior ahead of the MediaState
+ * refactor (Plan A, step 2), including known quirks that the refactor is
+ * intended to fix. When a quirk is resolved, the corresponding assertion here
+ * should change in the same PR.
+ */
+
+import { beforeEach, describe, expect, jest, test } from "@jest/globals";
+
+import { PipecatClient } from "../client";
+import {
+  DeviceError,
+  DeviceErrorType,
+  RTVIEvent,
+  RTVIMessageType,
+  TransportState,
+} from "../rtvi";
+import { TransportStub } from "./stubs/transport";
+
+/**
+ * Extends TransportStub with hooks needed for characterization:
+ * - count initDevices() calls
+ * - optionally throw on initDevices()
+ * - drive the onDeviceError callback directly
+ */
+class CharacterizationTransport extends TransportStub {
+  public initDevicesCallCount = 0;
+  public initDevicesShouldThrow: Error | undefined;
+
+  static override create(): CharacterizationTransport {
+    return new CharacterizationTransport();
+  }
+
+  public override initDevices(): Promise<void> {
+    this.initDevicesCallCount += 1;
+    if (this.initDevicesShouldThrow) {
+      const err = this.initDevicesShouldThrow;
+      // Match the shape of a real transport that begins enumeration and then
+      // fails partway — state flips to "initializing" before the rejection.
+      return new Promise<void>((_resolve, reject) => {
+        this.forceState("initializing");
+        setTimeout(() => reject(err), 10);
+      });
+    }
+    return super.initDevices();
+  }
+
+  private forceState(next: TransportState): void {
+    // The parent stub's state setter is private. Update the protected field
+    // directly and fire the registered callback so observers still see the
+    // transition.
+    this._state = next;
+    this._callbacks.onTransportStateChanged?.(next);
+  }
+
+  public emitDeviceError(error: DeviceError): void {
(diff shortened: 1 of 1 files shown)


client-js/tests/initDevices.spec.ts
```diff
@@ -0,0 +1,321 @@
...
+    // This is a baseline consumers rely on today: a device error does not
+    // drive TransportState. Plan A step 2 introduces a separate MediaState
+    // that IS updated from this event, leaving TransportState untouched.
+    expect(states).toEqual([]);
+    expect(client.state).toBe("disconnected");
+  });
+});
+
+describe("Transport contract: initial _state", () => {
+  test("abstract Transport initializes _state to 'disconnected'", () => {
```
mattieruth: im very confused about this test and the need for it. especially the last check on botready

mattieruth (commented):
this is great. smart to start with tests. i wish we didn't have to support the "initializing"/"initialized" states.

mattieruth (approved):

mattieruth:
A thought i just had of tests we should add:

1. initDevices() with cam/mic (either/both) disabled.
2. toggling cam/mic states before connect/after initDevices()
3. connect() with cam/mic disabled

client-js/tests/initDevices.spec.ts
Regaddi: Claude was honest about this 😄 

> Mattie's confusion is legitimate — that test is load-bearing in intent but sloppy in execution. Let me break it down:
> 
> What it's meant to lock in: the abstract Transport class declares protected _state: TransportState = "disconnected"; at client-js/client/transport.ts:33. That default value is THE sentinel the Plan A design calls out as carrying two meanings — "pre-init" and "post-disconnect." If someone changed that default (say, to "uninitialized" or "initial") the ambiguity would shift without the tests noticing. The first assertion (expect(transport.state).toBe("disconnected") before constructing the client) is the only line that exercises this — it reads the field straight off a freshly constructed transport, before initialize() has run.
> 
> Where I muddied it:
> 
> The client.state === "disconnected" assertion after new PipecatClient({ transport }) is redundant — the first test in the file ("initial TransportState is 'disconnected' (same sentinel as post-disconnect)") already covers that path.
> The expect(RTVIMessageType.BOT_READY).toBeDefined() line is a drive-by sanity check with no real connection to what the test is about. I wrote it to use an import I didn't otherwise need. It should go.
> Want me to tighten the test to just the abstract-class default assertion and drop the other two? That's the honest scope, and it'd read cleanly as "the abstract Transport's initial _state is the sentinel Plan A step 2 will disambiguate."

mattieruth:
i've also been trying to think if there's something to test around a scenario where both devices fail for different reasons, but Daily will only throw the error for one device so i'm not sure this is much different than what we have:  example: cam: 'not-found', mic: 'permissions' -- only the mic: 'permissions' error bubbles up. Of course, if you then try just a `setLocalVideo(true)`, you'll get the 'not-found' 🤔

Regaddi:
> i've also been trying to think if there's something to test around a scenario where both devices fail for different reasons, but Daily will only throw the error for one device so i'm not sure this is much different than what we have: example: cam: 'not-found', mic: 'permissions' -- only the mic: 'permissions' error bubbles up. Of course, if you then try just a `setLocalVideo(true)`, you'll get the 'not-found' 🤔

Let's put that into a separate follow-up. I'll shout in case this becomes a hurdle in the updated device state handling.

PR #203, opened 2026-04-25
Client updates for UI agent

markbackman:
## Summary

Adds the **UI Agent Protocol (v1)** to `@pipecat-ai/client-js` and `@pipecat-ai/client-react`. Lets server-side AI agents observe a GUI app via first-class `ui-snapshot` messages and drive it via `ui-command` messages, with first-class task lifecycle support. The matching server-side wire format lands in pipecat-ai/pipecat#4407; the higher-level Python `UIAgent` lives in `pipecat-ai-subagents`.

The wire format is **purely additive**: five new RTVI top-level message types (`ui-event`, `ui-command`, `ui-snapshot`, `ui-cancel-task`, `ui-task`) and a bump of `RTVI_PROTOCOL_VERSION` from `1.2.0` to `1.3.0`. Older 1.x endpoints continue to connect; they simply do not exercise the new UI message types.

## New API surface — `@pipecat-ai/client-js`

- `PipecatClient.sendUIEvent(event, payload?)` sends client-to-server `ui-event` data.
- `PipecatClient.startUISnapshotStream(options?)` and `stopUISnapshotStream()` manage streamed `{ tree: snapshot }` `ui-snapshot` data.
- `PipecatClient.cancelUITask(taskId, reason?)` asks the server to cancel an in-flight UI task group.
- `RTVIEvent.UICommand` and `RTVIEvent.UITask`, plus constructor callbacks `onUICommand` and `onUITask`, expose server-to-client protocol messages.
- `A11ySnapshotStreamer`, `snapshotDocument(root?, options?)`, and `findElementByRef(ref)` remain available for low-level DOM snapshot and command-targeting needs.
- Protocol data types live with the top-level RTVI message types: `UIEventData`, `UISnapshotData`, `UICommandData`, `UITaskData`, `UITaskListener`, and task lifecycle variants. UI command payload and a11y tree types remain in the UI protocol module.

## New API surface — `@pipecat-ai/client-react`

- `UITasksProvider` and `useUITasks()` expose in-flight UI task groups and cancellation.
- `useUIEventSender()` sends UI events from React components.
- `useUICommandHandler(command, handler)` registers a named UI command handler for the component lifetime.
- `useUISnapshot({ enabled, debounceMs, trackViewport, logSnapshots })` wraps `PipecatClient.startUISnapshotStream(...)`.
- Default DOM command handlers: `useDefaultScrollToHandler`, `useDefaultFocusHandler`, `useDefaultHighlightHandler`, `useDefaultSelectTextHandler`, `useDefaultSetInputValueHandler`, `useDefaultClickHandler`, and `useDefaultUICommandHandlers`.
- Typed sugar hooks: `useToastHandler(handler)` and `useNavigateHandler(handler)`.

React exports React-owned hooks, providers, and option/handler types. Protocol payload and wire data types are imported from `@pipecat-ai/client-js`.

## Test plan

- [ ] `npx jest` in `client-js/`
- [ ] `npx jest` in `client-react/`
- [ ] `npx tsc --noEmit` in both packages
- [ ] `npx eslint` in both packages
- [ ] End-to-end verified against the canonical server (`pipecat-ai-subagents` UIAgent) across the full command vocabulary in a working app


client-js/client/A11ySnapshotStreamer.ts
```diff
@@ -0,0 +1,248 @@
...
+ */
+
+/**
+ * Framework-agnostic helper that drives accessibility-snapshot
+ * streaming. Wraps the walker, a ``MutationObserver``, and the other
+ * triggers (scrollend, resize, focus, visibilitychange) into a single
+ * object with ``start()`` / ``stop()``. React apps use
+ * ``useA11ySnapshot`` which is a thin wrapper around this; vanilla JS
+ * or non-React apps instantiate the class directly.
+ */
```
Regaddi, line 14: The double backticks seem unintended. Probably worth cleaning up. There are a couple of comments with this.

client-js/client/A11ySnapshotStreamer.ts
mattieruth, line 1: file names should be camelCase or use underscores. generally, our convention should be that if a file primarily defines a class, then the file name should match the class (like we do in DailyReact). If it defines a suite of things, then it can be all lower_case_with_underscores (like rest_helpers). That said, I see we haven't been good at this in client-js since client.ts probably should've been PipecatClient.ts... so hmm... maybe:

- a11y-streamer.ts -> A11ySnapshotStreamer.ts or a11y_streamer.ts
- ui-agent-client -> UIAgentClient.ts or ui_agent_client.ts

WDYT?

client-js/client/client.ts
```diff
@@ -937,6 +950,63 @@ export class PipecatClient extends RTVIEventEmitter {
...
+  }
+
+  /**
+   * Start streaming accessibility snapshots to the server as
+   * first-class `ui-snapshot` RTVI messages.
+   *
+   * Calling this again replaces any existing managed streamer with
+   * the new options.
+   */
+  public startA11ySnapshotStream(
```
mattieruth: I feel like we don't need to specify A11y in our method name. We didn't in the RTVI protocol. How about `startUISnapshotStream()`/`stopUISnapshotStream()`? That also aligns it with the names of the other method in this suite of things.

client-js/client/client.ts
```diff
@@ -937,6 +950,63 @@ export class PipecatClient extends RTVIEventEmitter {
...
 
+  /**
+   * Send a named UI event to the server as a first-class RTVI
+   * `ui-event` message.
+   *
+   * @param event - App-defined event.
+   * @param payload - App-defined payload. Optional.
+   */
+  @transportReady
+  public sendUIEvent<T = unknown>(event: string, payload?: T): void {
```
mattieruth: did you ask claude about the necessity of all this fancy typing? `sendClientMessage()` is essentially an identical method without it.

client-js/rtvi/ui.ts
```diff
@@ -0,0 +1,343 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/**
+ * Shape of the `data` field of a `ui-event` RTVI message
+ * (client → server). `UIEventEnvelope` is what travels in the
+ * `data` field.
+ */
+export interface UIEventEnvelope<T = unknown> {
```
mattieruth: again. a question about the types. why this vs.
```
export type UIEventData = {
  event: string;
  payload: unknown;
};
```

also nit: i think the names here should be `Data` instead of `Envelope` for naming consistency with the types in `messages`. In fact, these top level types or maybe all of them should move there.

client-react/src/index.ts
```diff
@@ -68,3 +108,33 @@ export type {
   FunctionCallData,
   FunctionCallRenderer,
 } from "./conversation/types";
+
+// UI agent protocol re-exports from @pipecat-ai/client-js so React
+// consumers don't need a second import.
```
mattieruth: i'm confused about this. what's unique about theses types that we didn't have to do this for pre-existing types?

client-js/rtvi/ui.ts
```diff
@@ -0,0 +1,343 @@
...
+ */
+export interface UICommandEnvelope<T = unknown> {
+  /** App-defined command. */
+  command: string;
+  /** App-defined payload. */
+  payload: T;
+}
+
+/** Signature for a named UI command payload handler. */
+export type UICommandHandler<T = unknown> = (
```
mattieruth: is this type still necessary now that the command handling is done the same way as any other callback? i see it used in the `useUICommandHandler` but seems like a mismatch from how we handle all other callbacks but maybe it is truly unique. or if its only needed for react, maybe it should move there? (you're finding my knowledge gaps)

client-js/README.md
```diff
@@ -68,6 +68,47 @@ pcClient.on(RTVIEvent.Disconnected, () => {
 });
 ```
 
+## UI Agent Protocol (v1)
```
mattieruth, line 71: i'n torn on whether this really belongs here. the rest of the API isn't detailed in the README. I think I like the README being an install guide and quickstart and point to docs for everything else. But it also feels sacrilegious to delete documentation 😂

client-react/src/standardHandlers.ts
mattieruth, line 1: why the name `Standard` and not `UI` or `Command`? I'm trying to understand what the commonality across these are and "Standard" doesn't do it, unless it has a meaning in React I'm overlooking. Reading the comment below, maybe something like `defaultUICommandHandlers`

client-react/src/standardHandlers.ts
```diff
@@ -0,0 +1,472 @@
...
+   * sticky header. Positive ``top`` scrolls *up* by that amount so
+   * the target isn't hidden by a fixed header. Only applied when
+   * ``container`` is set (since window-level offsets are usually
+   * better solved with ``scroll-margin-top`` CSS).
+   */
+  offset?: { top?: number; left?: number };
+}
+
+/** Enable the default `scroll_to` handler: scrollIntoView on the target. */
+export const useStandardScrollToHandler = (
```
mattieruth: continuing on the naming topic... `useDefaultScrollToHandler` maybe?

client-js/client/A11ySnapshotStreamer.ts
markbackman: Fixed!

client-js/client/client.ts
markbackman: Fixed: the managed public APIs are now `startUISnapshotStream()` / `stopUISnapshotStream()`.

client-js/client/client.ts
markbackman: Fixed by simplifying `sendUIEvent` to a non-generic `UIEventData` shape with an `unknown` payload.

client-js/rtvi/ui.ts
markbackman: Fixed: the public wire shapes are now `Data` types, simplified to `unknown` payloads, and moved to `rtvi/messages.ts` with the other top-level RTVI message data types.

client-react/CHANGELOG.md
```diff
@@ -5,6 +5,35 @@ All notable changes to **Pipecat Client React** will be documented in this file.
...
+    `trackViewport`, `logSnapshots`.
+  - New `useUIEventSender()` and `useUICommandHandler(command, handler)`
+    hooks for the basics of the protocol.
+  - New opt-in standard command handlers:
+    `useStandardScrollToHandler({ block, inline, container, offset, ... })`,
+    `useStandardFocusHandler({ preventScroll })`,
+    `useStandardHighlightHandler({ className, defaultDurationMs, scrollIntoViewFirst })`,
+    plus `useStandardCommandHandlers(...)` to install all three at
+    once. Each resolves the target element by snapshot `ref` first,
+    then falls back to `document.getElementById(target_id)`.
```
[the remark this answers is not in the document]
markbackman: Fixed: the React changelog now names all six default command handlers and says the bundle installs all six.

client-js/rtvi/ui.ts
```diff
@@ -0,0 +1,343 @@
...
+  end_offset?: number;
+}
+
+/**
+ * Shape of the payload inside a `__ui_snapshot` UI event.
+ *
+ * A full tree is sent on each update; the server keeps the latest and
+ * renders it into `<ui_state>...</ui_state>` when an agent injects it.
+ */
+export interface A11ySnapshot {
```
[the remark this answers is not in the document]
markbackman: Fixed: the `A11ySnapshot` docs now describe first-class `ui-snapshot` messages with `{ tree: snapshot }` data.

client-js/rtvi/a11y_walker.ts
```diff
@@ -0,0 +1,765 @@
...
+const refToElement = new Map<string, WeakRef<Element>>();
+let refCounter = 0;
+
+function getRef(el: Element): string {
+  const existing = refMap.get(el);
+  if (existing) return existing;
+  const ref = `e${++refCounter}`;
+  refMap.set(el, ref);
+  refToElement.set(ref, new WeakRef(el));
+  return ref;
```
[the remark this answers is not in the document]
markbackman: Fixed: the ref registry now feature-detects `WeakRef` and falls back to strong refs with disconnected-element pruning.

client-js/rtvi/a11y_walker.ts
```diff
@@ -0,0 +1,765 @@
...
+ * a snapshot-known node. Walk up ``el.parentElement`` looking for the
+ * first ancestor that has a ref to find the closest snapshot-known
+ * container.
+ */
+export function findRefForElement(el: Element): string | null {
+  return refMap.get(el) ?? null;
+}
+
+/** Reset the ref counter and registry. Test-only helper. */
+export function __resetRefsForTesting(): void {
```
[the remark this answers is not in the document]
markbackman: Fixed: the public package entrypoint no longer re-exports `__resetRefsForTesting`; it remains directly available only for internal tests.

client-js/client/client.ts
```diff
@@ -937,6 +950,63 @@ export class PipecatClient extends RTVIEventEmitter {
...
+    options: A11ySnapshotStreamerOptions = {}
+  ): void {
+    this.stopA11ySnapshotStream();
+    this._a11ySnapshotStreamer = new A11ySnapshotStreamer((snapshot) => {
+      if (this.state !== "ready") return;
+      this._sendMessage(
+        new RTVIMessage(RTVIMessageType.UI_SNAPSHOT, { tree: snapshot })
+      );
+    }, options);
+    this._a11ySnapshotStreamer.start();
```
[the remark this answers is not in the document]
markbackman: Fixed: the client buffers the latest UI snapshot while the transport is not `ready` and flushes it immediately on the `ready` transition.

README.md
```diff
@@ -134,6 +134,18 @@ pcClient.on(RTVIEvent.Disconnected, () => {
...
 
+## UI Agent Protocol (v1)
+
+Both packages ship a v1 client implementation of the UI Agent Protocol, a structured wire format that lets server-side AI agents observe and drive a GUI app. Pairs with the `UIAgent` class in [`pipecat-subagents`](https://github.com/pipecat-ai/pipecat-subagents) on the Python side.
+
+The client side provides:
+
+- **`PipecatClient`** (`client-js`): sends `ui-event` messages, exposes `RTVIEvent.UICommand` / `RTVIEvent.UITask` subscriptions and matching constructor callbacks, manages accessibility snapshot streaming with `startA11ySnapshotStream(...)`, and can cancel UI task groups.
+- **`A11ySnapshotStreamer`** (`client-js`): low-level snapshot streaming implementation for advanced users.
+- **React bindings** (`client-react`): `UITasksProvider`, hooks (`useUIEventSender`, `useUICommandHandler`, `useUITasks`, `useA11ySnapshot`), and opt-in DOM defaults (`useStandardScrollToHandler`, `useStandardHighlightHandler`, `useStandardSelectTextHandler`, `useStandardSetInputValueHandler`, `useStandardClickHandler`, `useStandardFocusHandler`).
```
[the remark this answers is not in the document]
markbackman: Fixed: the PR description now matches the actual shipped API surface and removes stale `UIAgentClient` / `sendRTVIMessage` references.

client-react/src/index.ts
markbackman: Fixed: the broad React re-exports were removed so React keeps the existing package boundary. Protocol payload and wire data types should be imported from `@pipecat-ai/client-js`.

client-js/rtvi/ui.ts
markbackman: Fixed: `UICommandHandler` moved to React, where it is used to type the command hook; client-js now only owns the protocol data and payload types.

client-js/README.md
markbackman: Fixed: the package README section is trimmed back to a short pointer to the changelog and docs.

client-react/src/standardHandlers.ts
markbackman: Fixed: the file is now `defaultUICommandHandlers.ts`.

client-react/src/standardHandlers.ts
markbackman: Fixed: the exported hooks now use the `useDefault*Handler` / `useDefaultUICommandHandlers` names.

Regaddi (commented):
@markbackman I primarily skimmed through the React code and had a convo with Claude about the API, structure, usage patterns and I think this looks generally pretty good! The API surface seems fairly easy to understand, once you get the idea, and looks like a fairly thin wrapper around the client-js APIs.
I hope this feedback is structured simple and well enough to address the feedback items with ease.

### Summary

The React package is a clean, thin set of wrappers over the protocol primitives in `@pipecat-ai/client-js`. The split between wire-envelope types (in client-js) and reduced-state types (in client-react) is the right boundary, the snake_case → camelCase conversion at that boundary is intentional and correct, and `UseA11ySnapshotOptions extends A11ySnapshotStreamerOptions` is a good pattern to follow for any future `Use*Options` types.

Most feedback below is small. The two real correctness bugs are M1 (timer leak in `useStandardHighlightHandler`) and M2 (`aggregateStatus` has no `"running"` branch).

### Must-fix

**M1. `setTimeout` leak in `useStandardHighlightHandler`**
`client-react/src/standardHandlers.ts:206`

The timer ID is never stored or cleared. On unmount, option change, or rapid re-fires, the callback runs later and calls `classList.remove` on a stale element reference. Suggest storing the ID in a `useRef`, clearing it before setting a new one, and clearing it in the effect cleanup.

**M2. `aggregateStatus` silently maps `"running"` tasks to `"completed"`**
`client-react/src/UITasksProvider.tsx:17-22`

There is no branch for `"running"`. If `group_completed` arrives before one or more `task_completed` envelopes (reordering, dropped messages, server bug), the group flips to `"completed"` while sub-tasks still read `"running"`. A consumer rendering per-task status would show a completed group with running children. Suggest: if any task is still `"running"`, return `"running"` (or a `"unknown"` sentinel).

### Should-fix

**S1. `resolveTarget` calls `document.getElementById` unguarded — SSR risk on re-exports**
`client-react/src/standardHandlers.ts:59`

Inside command handlers it is safe (registration is effect-bound, never runs during SSR render). However, `findElementByRef` and `snapshotDocument` are re-exported from `index.ts` as live values. A consumer calling either during SSR (Next.js server component, `getServerSideProps`) hits a bare `document` reference. Suggest a `typeof document !== "undefined"` guard, plus a JSDoc note on the re-exported functions.

**S2. Test coverage gaps**

- No test files for `useStandardScrollToHandler`, `useStandardFocusHandler`, or `useStandardHighlightHandler` (`client-react/tests/`). The scroll-container/offset path, `preventScroll`, and the highlight class toggle (including the M1 timer leak) are unverified.
- `useStandardSelectTextHandler` does not cover inverted offsets on inputs (related to M3).
- `useA11ySnapshot` does not cover runtime `enabled: true → false → true` toggling, only the static-mount case. A `rerender` test would close this.
- `useUICommandHandler` and the silent-refusal paths in `useStandardSetInputValueHandler` / `useStandardClickHandler` (`disabled`, `aria-disabled`, `readonly`, hidden inputs) are not covered.

**S3. Silent refusals have no debug affordance**
`client-react/src/standardHandlers.ts` (`useStandardClickHandler`, `useStandardSetInputValueHandler`)

The intentional no-ops on `disabled` / `aria-disabled` / `readonly` / `type="hidden"` are the right behavior — the agent should not bypass affordances the user can't use. But there is no log, no callback, no way to diagnose "agent's click does nothing." Suggest at minimum a `logger.debug` line on each refusal path so developers can find these in dev tools when wiring things up. This could also be a dev-mode only logging (`process.env.NODE_ENV !== "production"`).

**S4. `useUITasks().groups` is append-only and unbounded over a session.**
`client-react/src/UITasksProvider.tsx`, `client-react/src/uiTasksTypes.ts`

Apps with long-lived sessions accumulate every task group ever dispatched, and the only way to render "recent tasks" is to slice in the render path. Two complementary additions would close this without breaking defaults:

1. Explicit dismissal on the API — mirrors how toast / notification systems work; the app knows when a group is "done" from the user's perspective and drops it.
```
// uiTasksTypes.ts
export interface UITasksAPI {
  groups: TaskGroup[];
  cancelTask: (taskId: string, reason?: string) => void;
  dismissTask: (taskId: string) => void;   // refused on running groups
  clearCompleted: () => void;
}
```
One reducer case each; trivial to implement.

2. Optional `maxGroups` cap on the provider — bounded by construction for apps that can't reasonably know when to dismiss (generic dashboards, debug UIs).
```
<UITasksProvider maxGroups={50}>
```
When the cap is exceeded, drop the oldest completed group; never drop running ones. Default unbounded so existing behavior is preserved.

The dismissal API is the load-bearing one — it's the standard pattern for this kind of state, gives apps full control, and avoids "where did my task go" surprises. The cap is the safety net for apps that don't have a natural dismissal point. Together they cover both shapes of consumer without imposing either.

Until then, the JSDoc on `useUITasks` should at least explicitly note the unbounded growth and recommend slicing in render.

**S5. `start > end` is not validated for `<input>` / `<textarea>` in `useStandardSelectTextHandler`**
`client-react/src/standardHandlers.ts:291-297`

The `Range` branch validates `start <= end` via the `inRange` check. The input/textarea branch passes offsets straight to `setSelectionRange`, which per spec clamps to a collapsed selection at `end` when `start > end` — the user sees no selection at all instead of the intended range. Suggest mirroring the `Range` branch's `inRange` check and falling back to `el.select()` on the invalid case (consistent with what the document path already does for invalid offsets).

### Nits / discussion

**N1. `ScrollBehavior` and `ScrollLogicalPosition` are redeclared**
`client-react/src/standardHandlers.ts:64-65`

Both are already in TypeScript's `lib.dom.d.ts` with identical shape. The local declarations shadow the built-ins. Suggest deleting both lines.

**N2. `useUICommandHandler` casts `payload as T` without narrowing**
`client-react/src/useUICommandHandler.ts:37`

Consistent with the design (callers own the type contract), but a malformed envelope will propagate to the handler silently. Suggest a JSDoc note that the cast is intentional and callers should validate if the payload is untrusted.

**N3. DOM glue could live in client-js**
The DOM logic in `useStandard*` handlers (text-node walking, container-relative scrolling, the disabled/readonly refusal rules) is genuinely React-agnostic. Moving it into client-js as plain functions, with the React package providing only `useUICommandHandler` registration, would let a vanilla / Vue / Svelte port reuse it without reimplementation. Architectural quibble, not a blocker.

**N4. `NO_OP_API` cast in `UITasksContext`**
`client-react/src/UITasksContext.ts:11-14`

`[] as TaskGroup[]` is harmless but unnecessary — the empty array literal already widens fine. Trivial.

### Notes for context (not action items)

- **Form-control gotcha worth documenting.** `set_input_value` covers text inputs, textareas, and `<select>` — but checkboxes and radios go through `click`. Developers coming from tools like Playwright might expect *dedicated* methods per control type (`fill` for text, `check`/`uncheck` for checkboxes, `selectOption` for selects), so collapsing checkbox/radio into the generic `click` command seems non-obvious. Worth calling out in the standard-handlers JSDoc and the README, alongside the existing `<select>` / combobox guidance.

client-react/src/standardHandlers.ts
markbackman: Fixed: the React DOM handler hooks now use the `useDefault...Handler` naming, with the bundle exposed as `useDefaultUICommandHandlers`.

client-js/client/A11ySnapshotStreamer.ts
markbackman: Cleaned up the double backticks.

markbackman:
## Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

mattieruth (approved):
You have resolved all my concerns. I grant you permission to proceed

Regaddi (commented):
Just a few behavioral observations worth double-checking compared to the last review run:

**Highlight handler is now single-active.** `useDefaultHighlightHandler` clears the previous highlight at the start of every invocation. Calling `highlight` on element B while A is still highlighted now removes A's class immediately, instead of letting both flash until each timer fires. For "walk the user through a sequence" UX this is arguably better; for "highlight these N items as a group" it no longer works (it always did, accidentally, before). If any agents in flight rely on simultaneous highlights, they will need to switch to a different mechanism.

**`select_text` on inverted offsets now selects everything.** Previously, `setSelectionRange(8, 3)` got spec-clamped to a collapsed cursor at position 3 (nothing visible). Now the `inRange` check fails and the handler falls back to `el.select()`, matching what the document-element branch already did. More consistent and probably what the agent intended, but it's a different observable result.

**`aggregateStatus` exposes stuck-task conditions.** With the new `"running"` branch in `aggregateStatus`, a group whose `group_completed` envelope arrives before all `task_completed` envelopes will stay `"running"` indefinitely instead of flipping to `"completed"`. This is the correct fix, but UI that gated a state transition on `group.status === "completed"` will now hold that transition forever in those cases instead of (incorrectly) advancing. Probably worth checking whether anything internal does that.

Regaddi (approved):

Regaddi:
> Just a few behavioral observations worth double-checking compared to the last review run:
> 
> **Highlight handler is now single-active.** `useDefaultHighlightHandler` clears the previous highlight at the start of every invocation. Calling `highlight` on element B while A is still highlighted now removes A's class immediately, instead of letting both flash until each timer fires. For "walk the user through a sequence" UX this is arguably better; for "highlight these N items as a group" it no longer works (it always did, accidentally, before). If any agents in flight rely on simultaneous highlights, they will need to switch to a different mechanism.
> 
> **`select_text` on inverted offsets now selects everything.** Previously, `setSelectionRange(8, 3)` got spec-clamped to a collapsed cursor at position 3 (nothing visible). Now the `inRange` check fails and the handler falls back to `el.select()`, matching what the document-element branch already did. More consistent and probably what the agent intended, but it's a different observable result.
> 
> **`aggregateStatus` exposes stuck-task conditions.** With the new `"running"` branch in `aggregateStatus`, a group whose `group_completed` envelope arrives before all `task_completed` envelopes will stay `"running"` indefinitely instead of flipping to `"completed"`. This is the correct fix, but UI that gated a state transition on `group.status === "completed"` will now hold that transition forever in those cases instead of (incorrectly) advancing. Probably worth checking whether anything internal does that.

These can be addressed in follow-ups, if necessary 🙌

PR #204, opened 2026-04-28
feat(js): add MediaState lifecycle on PipecatClient

Regaddi:
## Summary

Plan A step 2. Introduces a per-device `MediaState` (`mic`, `cam`) on `PipecatClient`, independent of `TransportState`. Resolves the `'disconnected'` sentinel ambiguity that step 1 characterized: consumers now have a separate, unambiguous signal for device readiness vs session state.

Modelled after [daily-react's DailyDevices](https://github.com/daily-co/daily-react/blob/main/src/DailyDevices.tsx). All additive, minor-version. No breaking changes.

## New API surface

- `DeviceState` union: `"uninitialized" | "initializing" | "granted" | "error"`.
- `DeviceErrorReason` union: `"blocked" | "already-in-use" | "not-found" | "invalid-constraints" | "not-supported" | "unknown"`.
- `DeviceStatus` discriminated union: `{ state: 'uninitialized' | 'initializing' | 'granted' }` or `{ state: 'error', reason: DeviceErrorReason, details?: unknown }`. Consumers branch on `state` and TS narrows; error info no longer requires checking against an exclusion list of "non-error" strings.
- `MediaState` type: `{ mic: DeviceStatus, cam: DeviceStatus }` (speakers intentionally excluded — no permission/in-use mode).
- `RTVIEvent.MediaStateUpdated` event.
- `onMediaStateChanged` callback in `RTVIEventCallbacks`.
- `client.mediaState` getter (returns a deep snapshot).
- `client.needsInit()` public method — same logic that drives the implicit-init gate, exposed for consumers (e.g. step 3's hooks) to branch on.

## Behavior

- `initDevices()` drives both devices into `'initializing'` on entry. Per-device `'granted'` transitions are sourced from `onMicUpdated` / `onCamUpdated` events the transport fires with a real `deviceId`. **MediaState mirrors what the transport actually did, not what the caller asked for** — necessary because real transports are asymmetric (e.g. daily-js's `startCamera` honors `startVideoOff` but acquires the mic regardless of `startAudioOff`).
- Devices the transport never reports acquired fall back to `'uninitialized'` on success, `{ state: 'error', reason: 'unknown' }` on failure.
- `DeviceError` events update MediaState per-device via a daily-react-style classifier. Speaker-only errors are no-ops for MediaState. Devices in error states (e.g. `not-found` because the cam was unplugged) can recover to `'granted'` on a subsequent `initDevices()` call when the device becomes available again.
- The implicit-init gate in `connect()` and `startBot()` is now `needsInit()` instead of `state === 'disconnected'`. **Reconnect no longer double-fires `initDevices()`**, and pickers stay accessible after a disconnect (the contract step 4 voice-ui-kit will rely on). The gate considers only opted-in devices, so callers who set both `enableMic` and `enableCam` to `false` don't hit a redundant init loop.
- **Permissions API enrichment runs post-transport.** A `'denied'` answer from `navigator.permissions.query({ name })` authoritatively flips the device to `{ state: 'error', reason: 'blocked' }`, overriding any under-reported `DeviceError` from the transport (concrete case: on a page where permissions were previously blocked, daily-js's `camera-error` only names whichever device was tried first — the post-transport re-query catches the missing one). Silent fallback when the API is unavailable or throws (Safari, mobile browsers). Worth a daily-js follow-up ticket.

## Decisions locked in

These were deferred in the design doc; resolved here so step 3 / step 4 don't re-litigate:

1. **Permissions enrichment is post-transport only** and authoritative for `'denied'`. Pre-transport enrichment was dropped — the timing benefit is imperceptible vs the post-transport flip.
2. **Pre-session auto-trigger stays opt-in.** `PipecatClient` does not auto-call `initDevices()`. The current behavior (explicit `initDevices()`, or implicit via `connect()` / `startBot()`) is preserved. Step 3's state provider will expose a prop for consumers that want auto-init on mount, defaulting off.
3. **Transport idempotency handled at the client layer** via `needsInit`. No transport changes required. Plan B3 can revisit at the transport level later.
4. **Recovery is `initDevices()` itself.** A second call re-enters `'initializing'` and reclassifies. Per-device upgrades to `'granted'` are not gated on prior state — a previously errored device can recover when conditions change. No separate `retryDevices()` method.
5. **`enableMic` / `enableCam` are hints to the transport, not drivers of MediaState.** Underlying transports may acquire devices the caller didn't opt into (or skip ones they did). The user-visible state always reflects what's actually happening rather than what was requested.

## Tests

- `tests/mediaState.spec.ts` (29 tests): lifecycle, classification matrix (one test per `DeviceErrorType` mapped to its `DeviceErrorReason`), partial-error scenarios, recovery from error states, `needsInit` gate (including the public method exposure), Permissions API enrichment, and "transport didn't acquire" cases. Uses `UNINIT` / `INITIALIZING` / `GRANTED` / `errored(reason)` shorthand for readability.
- Step-1 reconnect test updated to assert the new no-double-init behavior and check `mediaState` directly. Both test stubs (`MediaStateTransport`, `CharacterizationTransport`) fire `onMicUpdated` / `onCamUpdated` after `super.initDevices()` resolves, mirroring real transport timing.
- Full suite: **110 / 110 pass**.

## Test plan

- [x] `npx jest` — 110 / 110 pass
- [x] `npx tsc --noEmit` — clean
- [x] `npx eslint client/ rtvi/ tests/` — clean
- [x] Hand-tested in a Vite playground against `SmallWebRTCTransport` across the four `enableMic`/`enableCam` combinations and several permission-block paths. Playground intentionally not committed.

### DIFF
--- a/client-js/client/client.ts
+++ b/client-js/client/client.ts
@@ -15,6 +15,8 @@ import {
   BotReadyData,
   BotTTSTextData,
   ClientMessageData,
+  DeviceErrorReason,
+  DeviceStatus,
   ErrorData,
   LLMContextMessage,
   LLMFunctionCallData,
@@ -23,6 +25,7 @@ import {
   LLMFunctionCallResultResponse,
   LLMFunctionCallStartedData,
   LLMFunctionCallStoppedData,
+  MediaState,
   Participant,
   PipecatMetricsData,
   RTVIEvent,
@@ -57,6 +60,31 @@ export type FunctionCallParams = {
   arguments: Record<string, unknown>;
 };
 
+/**
+ * Map a DeviceErrorType onto a DeviceErrorReason that captures the per-device
+ * failure mode. Speaker-affecting errors are not represented here because
+ * MediaState only tracks mic and cam.
+ */
+function deviceErrorReasonFromType(
+  type: RTVIErrors.DeviceErrorType
+): DeviceErrorReason {
+  switch (type) {
+    case "in-use":
+      return "already-in-use";
+    case "permissions":
+      return "blocked";
+    case "not-found":
+      return "not-found";
+    case "undefined-mediadevices":
+      return "not-supported";
+    case "constraints":
+      return "invalid-constraints";
+    case "unknown":
+    default:
+      return "unknown";
+  }
+}
+
 export type FunctionCallCallback = (
   fn: FunctionCallParams
 ) => Promise<LLMFunctionCallResult | void>;
@@ -87,6 +115,7 @@ export type RTVIEventCallbacks = Partial<{
   onMicUpdated: (mic: MediaDeviceInfo) => void;
   onSpeakerUpdated: (speaker: MediaDeviceInfo) => void;
   onDeviceError: (error: RTVIErrors.DeviceError) => void;
+  onMediaStateChanged: (mediaState: MediaState) => void;
   onTrackStarted: (track: MediaStreamTrack, participant?: Participant) => void;
   onTrackStopped: (track: MediaStreamTrack, participant?: Participant) => void;
   onScreenTrackStarted: (
@@ -182,6 +211,14 @@ export class PipecatClient extends RTVIEventEmitter {
   private _botTranscriptionWarned = false;
   private _llmFunctionCallWarned = false;
 
+  // Per-device device state. Independent of TransportState — driven by
+  // initDevices() and DeviceError events, never by transport connect/disconnect.
+  // See common_types.ts for the rationale and the daily-react reference.
+  private _mediaState: MediaState = {
+    mic: { state: "uninitialized" },
+    cam: { state: "uninitialized" },
+  };
+
   constructor(options: PipecatClientOptions) {
     super();
 
@@ -275,10 +312,14 @@ export class PipecatClient extends RTVIEventEmitter {
         this.emit(RTVIEvent.AvailableSpeakersUpdated, speakers);
       },
       onCamUpdated: (cam) => {
+        // Real device selected → permission was granted. Upgrade MediaState
+        // (no-op if already granted, sticky if blocked / in-use / etc).
+        if (cam?.deviceId) this._markDeviceGranted("cam");
         options?.callbacks?.onCamUpdated?.(cam);
         this.emit(RTVIEvent.CamUpdated, cam);
       },
       onMicUpdated: (mic) => {
+        if (mic?.deviceId) this._markDeviceGranted("mic");
         options?.callbacks?.onMicUpdated?.(mic);
         this.emit(RTVIEvent.MicUpdated, mic);
       },
@@ -287,6 +328,12 @@ export class PipecatClient extends RTVIEventEmitter {
         this.emit(RTVIEvent.SpeakerUpdated, speaker);
       },
       onDeviceError: (error) => {
+        // Classify into MediaState in real time. Works the same whether the
+        // error fires during an initDevices() call (mid-await) or out of band
+        // (e.g. devicechange-driven). The post-transport Permissions API
+        // re-query in initDevices() then has the final word for any 'denied'
+        // override.
+        this._classifyAndApplyDeviceError(error);
         options?.callbacks?.onDeviceError?.(error);
         this.emit(RTVIEvent.DeviceError, error);
       },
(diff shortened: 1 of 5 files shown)


client-js/client/client.ts
```diff
@@ -412,11 +483,108 @@ export class PipecatClient extends RTVIEventEmitter {
   // ------ Transport methods
 
   /**
-   * Initialize local media devices
+   * Initialize local media devices.
+   *
+   * Drives MediaState transitions: both mic and cam move to 'initializing' on
+   * entry, then to 'granted' on success. On failure the in-flight DeviceError
+   * (if any) classifies them per-device; otherwise both fall back to
+   * 'unknown'. The original error is always re-thrown.
```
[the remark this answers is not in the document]
Regaddi: Applied. The docstring now reads:

```ts
/**
 * Initialize local media devices.
 *
 * Drives MediaState transitions: both mic and cam move to 'initializing' on
 * entry. On success, each device moves to 'granted' only if the transport
 * reports it as acquired (via onMicUpdated / onCamUpdated with a real
 * deviceId); otherwise that device falls back to 'uninitialized'. On
 * failure the in-flight DeviceError (if any) classifies the affected
 * device(s) per-device; anything still at 'initializing' falls back to
 * 'unknown'. The original error is always re-thrown.
 *
 * Calling this again after a failure is the recovery path — a second call
 * re-enters 'initializing' and reclassifies. There is no separate
 * retryDevices() method.
 */
```

client-js/rtvi/common_types.ts
```diff
@@ -28,3 +28,40 @@ export type Participant = {
...
+  /** initDevices() is in flight (enumeration / permission request) */
+  | "initializing"
+  /** Permission prompt is visible to the user */
+  | "pending"
+  /** Device is enumerated and ready to use */
+  | "granted"
+  /** User explicitly denied permission */
+  | "blocked"
+  /** Hardware is busy in another application */
+  | "in-use"
```
mattieruth: when i read through these states without the comments, i had expected this to mean that it was setup and being used by THIS application. So I might suggest `already-in-use` to better indicate an error state.

client-js/client/client.ts
```diff
@@ -57,6 +59,52 @@ export type FunctionCallParams = {
...
+  switch (type) {
+    case "in-use":
+      return "in-use";
+    case "permissions":
+      return "blocked";
+    case "not-found":
+      return "not-found";
+    case "undefined-mediadevices":
+      return "not-supported";
+    case "constraints":
```
mattieruth, line 80: i don't think this should be bucketed as "unknown". It's definitely known and easy to hit. maybe 'invalid-constraints' or 'bad-constraints'

client-js/client/client.ts
```diff
@@ -412,11 +483,111 @@ export class PipecatClient extends RTVIEventEmitter {
...
+
+  /**
+   * Upgrade a device from 'initializing' / 'uninitialized' to 'granted'.
+   * Called from the wrapped onMicUpdated / onCamUpdated when the transport
+   * reports an actual selected device. Other states (blocked, in-use, etc.)
+   * are sticky until the next initDevices() call resets them.
+   */
+  private _markDeviceGranted(kind: "mic" | "cam"): void {
+    const current = this._mediaState[kind];
+    if (current === "initializing" || current === "uninitialized") {
```
mattieruth: i don't think you want to gate this on the prior state. It's completely possible for an error state to turn into a non-error state: example: starting with a camera not plugged in, getting an error. Then plugging it in and succeeding.

client-js/rtvi/common_types.ts
```diff
@@ -28,3 +28,40 @@ export type Participant = {
...
+ * explicit calls to PipecatClient.initDevices() and by DeviceError events.
+ * Never by transport connect/disconnect.
+ */
+export type DeviceStatus =
+  /** initDevices() has not been called yet for this device */
+  | "uninitialized"
+  /** initDevices() is in flight (enumeration / permission request) */
+  | "initializing"
+  /** Permission prompt is visible to the user */
+  | "pending"
```
mattieruth: how do you plan to support this state? i don't believe you can determine this from the browser and daily does not expose it.

client-js/rtvi/common_types.ts
```diff
@@ -28,3 +28,40 @@ export type Participant = {
...
 };
+
+/**
+ * Per-device lifecycle status, tracked independently for the mic and the cam.
+ *
+ * Modelled after daily-react's DailyDevices state. Transitions are driven by
+ * explicit calls to PipecatClient.initDevices() and by DeviceError events.
+ * Never by transport connect/disconnect.
+ */
+export type DeviceStatus =
```
mattieruth: i'm questioning whether DeviceStatus is sufficiently covered by a single string. We surface so much more in errors. It's also not ideal to determine error state by checking if the state is not one of n states vs. just checking for "error".  What do you think of making DeviceStatus something like:

```
export type DeviceState = 
  /** initDevices() has not been called yet for this device */
  | "uninitialized"
  /** initDevices() is in flight (enumeration / permission request) */
  | "initializing"
  /** Device is enumerated and ready to use */
  | "granted"
  /** Device is in an error state and could not be acquired */
  | "error";

export type DeviceStatus = {
  state: DeviceState;
};

export type DeviceErrorTypes = 
  /** User explicitly denied permission */
  | "blocked"
  /** Hardware is busy in another application */
  | "in-use"
  ...;

export type DeviceErrorStatus extends DeviceStatus {
  error: DeviceErrorTypes;
  details?: unknown // (or we type the possibilities out)
}
```

client-js/tests/initDevices.spec.ts
```diff
@@ -46,6 +46,15 @@ class CharacterizationTransport extends TransportStub {
...
     }
+    // Mirror what a happy-path transport does: report selected devices.
+    // PipecatClient watches onMicUpdated/onCamUpdated to upgrade MediaState
+    // to 'granted', which is what the needsInit gate keys off of.
+    this._callbacks.onMicUpdated?.({
+      deviceId: "mic-default",
+    } as MediaDeviceInfo);
+    this._callbacks.onCamUpdated?.({
+      deviceId: "cam-default",
+    } as MediaDeviceInfo);
```
mattieruth: if we wanted to be more accurate to how events come through, I think these should come after the `super.initDevices()`;

client-js/client/client.ts
```diff
@@ -412,11 +483,111 @@ export class PipecatClient extends RTVIEventEmitter {
...
+      await this._transport.initDevices();
+      // Transport resolved. Devices granted permission have already been
+      // upgraded by onMicUpdated / onCamUpdated event handlers during the
+      // await. Anything still at 'initializing' is a device the transport
+      // didn't speak to — fall back to 'uninitialized'. Then overlay any
+      // in-flight DeviceError so the affected device(s) carry the right
+      // status.
+      this._resolveLingeringInitializing("uninitialized");
+      if (this._pendingDeviceError) {
+        this._classifyAndApplyDeviceError(this._pendingDeviceError);
```
mattieruth: i'm confused about this logic (or really, the need for `_pendingDeviceError` at all). It looks like the only thing that sets it is `onDeviceError`, where `this._classifyAndApplyDeviceError` is already called with the error, so if `this._pendingDeviceError` is not null, i think we can assume this has already done. The only other place `_pendingDeviceError` is used is in `_resolveLingeringInitializing`, see note there...

client-js/client/client.ts
```diff
@@ -412,11 +483,111 @@ export class PipecatClient extends RTVIEventEmitter {
...
+   * onMicUpdated / onCamUpdated and isn't named by an in-flight DeviceError.
+   * Apply the supplied fallback so it doesn't linger in 'initializing'.
+   *
+   * On success, fallback is 'uninitialized' (the transport simply didn't
+   * speak to that device — e.g. daily-js skipping cam under
+   * startVideoOff: true). On failure, fallback is 'unknown' (we know
+   * something went wrong but can't pin it on this device).
+   */
+  private _resolveLingeringInitializing(fallback: DeviceStatus): void {
+    const flagged = this._pendingDeviceError?.devices ?? [];
```
mattieruth: i don't think we need to check against the error here. If there was an error on the device, then `this._mediaState[kind]` by definition shouldn't be 'initializing'

client-js/client/client.ts
```diff
@@ -57,6 +59,52 @@ export type FunctionCallParams = {
...
+/**
+ * True when initDevices() still has work to do. Returns true if any device
+ * the caller opted into (enableMic / enableCam) is still 'uninitialized'.
+ * Devices the caller opted out of are not considered — they stay
+ * 'uninitialized' by design and must not gate the implicit init.
+ *
+ * Used by connect() and startBot(). Replaces the historical
+ * `state === "disconnected"` gate, which conflated pre-init and post-session.
+ */
+function needsInit(
```
mattieruth: i'm curious why this isn't a member function. seems like it could be handy for clients to be able to call themselves. Plus, the only times we use it, we always pass in (this._mediaState, this._options), it's never used with just a generic state or set of options.

client-js/client/client.ts
```diff
@@ -412,11 +483,111 @@ export class PipecatClient extends RTVIEventEmitter {
...
+      this._resolveLingeringInitializing("unknown");
+      if (this._pendingDeviceError) {
+        this._classifyAndApplyDeviceError(this._pendingDeviceError);
+      }
+      throw error;
+    } finally {
+      // Re-query the Permissions API now that the prompt (if any) has been
+      // dismissed. Some transports under-report blocked devices in their
+      // DeviceError (e.g. a combined prompt for both mic+cam blocked, but the
+      // transport only names one). Permissions API is the authoritative
```
mattieruth: this comment isn't quite right. a combined prompt with one device disabled will report both as blocked in the camera-error. It's only when you start the cam or mic and permissions were previously granted or blocked (so no permission prompt shown) that the camera-error will overlook a blocked state.  (side note: we should make a linear ticket to consider fixing this is daily :) )

example:
```
// on a page with client setup like so:
client = new PipecatClient({
  enableMic: false,
  enableCam: true,
  transport,
  callbacks: {
    onDeviceError: (e) => { console.error('device error. devices affected', e.devices) }
  }
});

// load page first time ever
client.initDevices(); // will show permission prompt for mic & cam
// user rejects permissions
// console will log: 'device error. devices affected ["cam", "mic"]
// reload page
client.initDevices(); // no prompt shown
// console will log: 'device error. devices affected ["cam"]
```

so doing the enrichment is sound, but the comment is misleading.

client-js/client/client.ts
```diff
@@ -412,11 +483,111 @@ export class PipecatClient extends RTVIEventEmitter {
...
+    // deviceId only when permission was granted. Devices the transport never
+    // speaks to fall back to 'uninitialized' below.
+    this._setMediaState({ mic: "initializing", cam: "initializing" });
+
+    // Optimistic permissions enrichment (daily-react pattern). If the API is
+    // available and either device is already 'denied', flip it to 'blocked'
+    // so consumers can render the right UI before the underlying transport
+    // call fails. Wrapped in try/catch — `permissions.query` is missing or
+    // partial on Safari and several mobile browsers.
+    await this._enrichFromPermissionsAPI();
```
mattieruth: i don't really get the need for this being here. I could see maybe putting this in the constructor to catch cases where permissions are blocked right out of the gate, but having it here, the timing difference of reporting the device blocked here vs as part of initDevices will be almost imperceptible.

client-js/client/client.ts
Regaddi: Went with `"invalid-constraints"` 👍

client-js/client/client.ts
Regaddi: Good call! Updated!

client-js/rtvi/common_types.ts
Regaddi: 😬 Removed!

client-js/rtvi/common_types.ts
Regaddi: Agreed, this seems more fine-grained 👍

client-js/client/client.ts
Regaddi: Promoted ☺️

mattieruth (approved):
you really took my review to heart 😍 . I think this looks a lot better. I didn't do a complete re-review, just checked all the changes from yesterday and they look great. Thanks for taking this on!

PR #205, opened 2026-05-04
feat(react): add useMediaState hook + autoInitDevices opt-in

Regaddi:
## Summary

Plan A step 3. Builds on the `MediaState` surface added in client-js 1.8.0 (#204) to give React consumers a per-device device-state hook and the opt-in auto-init contract that step 2 deferred here.

All additive. Existing hook return shapes unchanged.

## Architecture: centralized state ownership

A new `PipecatClientMediaStateProvider` (internal, rendered automatically by `PipecatClientProvider`) owns the jotai atom that mirrors `client.mediaState`. It:

- **Seeds** the atom from `client.mediaState` on mount and on client changes.
- **Subscribes** to `RTVIEvent.MediaStateUpdated` and writes incoming payloads into the atom.

`useMediaState` only reads from the atom — no per-hook `useEffect`, no per-hook seed. A hook that mounts after the events have already fired (e.g. a settings page navigated to mid-session) sees the current state immediately because the provider has been keeping the atom in sync the whole time.

This matches the existing `PipecatClientStateProvider` pattern (rendered internally by `PipecatClientProvider`, not exported) at the cost of one extra nesting layer in the component tree.

### `useMediaState()`

```ts
const { mic, cam } = useMediaState();
// mic.state: 'uninitialized' | 'initializing' | 'granted' | 'error'
// when state === 'error': mic.reason: 'blocked' | 'already-in-use' | ...
```

### `autoInitDevices` on `PipecatClientProvider`

```tsx
<PipecatClientProvider client={client} autoInitDevices>
  ...
</PipecatClientProvider>
```

When `true`, the provider calls `client.initDevices()` on mount if `client.needsInit()` returns true. **Defaults to `false`** to preserve the safe "no preemptive permission prompts" behavior — matches the [step 2 decision](https://github.com/pipecat-ai/pipecat-client-web/pull/204) we locked in. Apps that already gathered consent (e.g. via signup or a "join" button) opt in; everyone else keeps today's behavior.

Errors during auto-init are intentionally swallowed at the provider — consumers observe failures via `RTVIEvent.DeviceError`, `RTVIEvent.MediaStateUpdated`, or by reading `client.mediaState` / `useMediaState()`.

## Notes on what's _not_ in this PR

The Notion plan also called for an `isReady` convenience on `usePipecatClientMediaDevices()`. **Dropped:** the meaning of "ready" is app-specific — a voice-only app considers itself ready when mic is `granted`, a voice+video app needs both, an app where cam is optional has its own definition. The core library should return technical truth via `useMediaState()`; consumers compute their own readiness predicate. Examples:

```ts
// Voice-only
const isReady = mic.state === "granted";

// Voice + video
const isReady = mic.state === "granted" && cam.state === "granted";

// Voice + optional video
const isReady = mic.state === "granted";
const hasVideo = cam.state === "granted";
```

## Tests

No new tests in this PR. Hooks in `client-react/src/` have no existing tests today (`usePipecatClientMediaDevices`, `usePipecatClientCamControl`, etc.). Adding hook-test infrastructure (`@testing-library/react` + a provider harness) is a meaningful infra investment that should be its own PR. The new code is a thin wrapper around well-tested machinery in client-js (#204 added 110 tests covering MediaState semantics).

Manual smoke-tested against `SmallWebRTCTransport` in a Vite playground covering:
- `useMediaState()` reflects per-device transitions live.
- A consumer that mounts after `initDevices()` already ran sees the current state immediately (the central seed in the new provider).
- `autoInitDevices={true}` triggers `initDevices()` on mount; `false` does not.
- Idempotency: remounting the provider with `autoInitDevices={true}` doesn't re-trigger when devices are already granted (`needsInit()` short-circuits).

## Test plan

- [x] `npx tsc --noEmit -p client-react` — clean
- [x] `npx eslint client-react/src/` — clean (the 700+ workspace-level errors are pre-existing test-file `no-undef` issues unrelated to this change)
- [x] `npx jest -w @pipecat-ai/client-react` — 185 / 185 pass (no new tests; existing suite still green)
- [x] Manual verification per the list above

## Step 4 hand-off

The contract `voice-ui-kit`'s `UserAudioControl` / `UserVideoControl` will use:

- Hide spinner / show picker when the relevant device is `state: 'granted'`.
- Show a "permission denied" affordance when `state: 'error', reason: 'blocked'`.
- Show "no hardware" / "device busy" / etc. for the remaining error reasons.
- Keep the picker accessible across `connect` / `disconnect` cycles — `mediaState` no longer regresses to `uninitialized` post-session.

### DIFF
--- a/client-react/src/PipecatClientMediaState.tsx
+++ b/client-react/src/PipecatClientMediaState.tsx
@@ -1,0 +1,78 @@
+/**
+ * Copyright (c) 2026, Daily.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+import { MediaState, RTVIEvent } from "@pipecat-ai/client-js";
+import { atom, useAtomValue } from "jotai";
+import { useAtomCallback } from "jotai/utils";
+import React, { useCallback, useEffect } from "react";
+
+import { usePipecatClient } from "./usePipecatClient";
+import { useRTVIClientEvent } from "./useRTVIClientEvent";
+
+const DEFAULT_MEDIA_STATE: MediaState = {
+  mic: { state: "uninitialized" },
+  cam: { state: "uninitialized" },
+};
+
+/**
+ * Module-scoped jotai atom holding the current per-device MediaState.
+ * Owned by PipecatClientMediaStateProvider — that's the single point that
+ * seeds it from `client.mediaState` and writes to it on every
+ * RTVIEvent.MediaStateUpdated. Hooks (useMediaState,
+ * usePipecatClientMediaDevices) only read.
+ */
+const mediaStateAtom = atom<MediaState>(DEFAULT_MEDIA_STATE);
+
+/**
+ * Provider that mirrors the underlying PipecatClient's MediaState into the
+ * shared jotai atom. Rendered automatically by PipecatClientProvider; not
+ * exported, so apps don't need to wire it up themselves.
+ *
+ * Centralizing the subscription here means consumer hooks stay thin — they
+ * just read the atom, with no per-hook seed effect — and a hook that mounts
+ * after initDevices() has run still sees the current state immediately.
+ */
+export const PipecatClientMediaStateProvider: React.FC<
+  React.PropsWithChildren
+> = ({ children }) => {
+  const client = usePipecatClient();
+
+  const setMediaState = useAtomCallback(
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
LGTM!

PR #211, opened 2026-05-22
fix: respect final flag in mergeMessages and handle user-llm-text event

### LINKED ISSUE #210: Two conversation bugs: mergeMessages ignores final flag + user-llm-text has no RTVIEvent mapping
## Summary

Two related bugs in `@pipecat-ai/client-js` + `@pipecat-ai/client-react` (reproduced on client-js 1.8.0 / client-react 1.4.0 / pipecat server 1.2.1).

---

### Description

`mergeMessages` (in `client-react/src/conversation/conversationActions.ts`) merges consecutive same-role messages within a 30-second window. The merge fires even when the previous assistant message has already been marked `final: true` by `bot-stopped-speaking`. This causes text from a completed bot turn to accumulate into the next turn's bubble.

### Steps to Reproduce

1. Use any pipeline where bot responses are short enough that two turns occur within 30 seconds.
2. Send a message — bot responds with turn 1 (e.g. `"Hello! How can I help you today?"`).
3. Send a second message immediately — bot responds with turn 2 (e.g. `"I can help you with many tasks."`).

**Expected transcript:**
```
User:  Hello
Bot:   Hello! How can I help you today?
User:  What can you do?
Bot:   I can help you with many tasks.
```

**Actual transcript:**
```
User:  Hello
Bot:   Hello! How can I help you today?
User:  What can you do?
Bot:   Hello! How can I help you today? I can help you with many tasks.
```

Turn 1's full text is prepended to turn 2's bubble.

### Root Cause

In `conversationActions.ts`, the `shouldMerge` condition:

```ts
const shouldMerge =
  lastMerged &&
  lastMerged.role === currentMessage.role &&
  currentMessage.role !== "system" &&
  currentMessage.role !== "function_call" &&
  timeDiff < MERGE_WINDOW_MS;
```

does **not** check `lastMerged.final`. A finalized message is treated identically to an in-progress one, so turn 2's parts get concatenated onto turn 1's parts array.

### Suggested Fix

Add a `final` guard to the `shouldMerge` condition:

```ts
const shouldMerge =
  lastMerged &&
  lastMerged.role === currentMessage.role &&
  currentMessage.role !== "system" &&
  currentMessage.role !== "function_call" &&
  !lastMerged.final &&   // ← add this
  timeDiff < MERGE_WINDOW_MS;
```

---

### Description

In chat mode, the server emits `user-llm-text` instead of `user-transcription` (text input bypasses STT entirely). The client SDK does not map `user-llm-text` to any `RTVIEvent`, and `usePipecatConversation` does not handle it. As a result, user messages are never added to the conversation array in chat mode. This compounds Bug 1: with no user messages acting as separators, every bot turn is adjacent to the previous assistant entry and `mergeMessages` always fires.

### Root Cause

- `RTVIMessageType.USER_LLM_TEXT = "user-llm-text"` is defined in `client-js/rtvi/messages.ts`
- But `RTVIEvent` in `client-js/rtvi/events.ts` has no corresponding `UserLlmText` entry
- `PipecatClient.handleMessage()` in `client-js/client/client.ts` has no `case` for `RTVIMessageType.USER_LLM_TEXT`, so messages are silently dropped
- `useConversationEventWiring` (client-react) has no handler for user LLM text events

Note: `user-llm-text` also fires in voice mode (when the STT result reaches the LLM), making it a unified "user input reached LLM" event across both modes.

### Suggested Fix

**In `client-js/rtvi/events.ts`**, add to `RTVIEvent`:

```ts
UserLlmText = "userLlmText",
```

**In `client-js/client/client.ts`**, add a case in `handleMessage`:

```ts
case RTVIMessageType.USER_LLM_TEXT:
  this._options.callbacks?.onUserLlmText?.(ev.data);
  this.emit(RTVIEvent.UserLlmText, ev.data);
  break;
```

**In `client-react/src/conversation/useConversationEventWiring.ts`**, wire the event:

```ts
client.on(RTVIEvent.UserLlmText, ({ text }) => {
  applyUserTranscriptText(get, set, text, true);
});
```

---

## Environment

| Package | Version |
|---|---|
| `@pipecat-ai/client-js` | 1.8.0 |
| `@pipecat-ai/client-react` | 1.4.0 |
| pipecat server | 1.2.1 |
| Transport | LiveKit (custom) |
| Pipeline | DummySTT + DummyTTS (chat mode) |

A PR with minimal fixes for both bugs is being submitted alongside this issue.

rahulsolanki001:
Fixes #210

## Summary

- **`mergeMessages` final guard**: `shouldMerge` in `conversationActions.ts` now checks `!lastMerged.final`, so a finalized bot turn is never merged with the next turn's messages regardless of the 30-second window.
- **`user-llm-text` event wiring**: Added `RTVIEvent.UserLlmText` to the event enum, a `USER_LLM_TEXT` case in `PipecatClient.handleMessage`, `onUserLlmText` to the callbacks type, and a handler in `useConversationEventWiring` that calls `upsertUserTranscript` — making chat-mode user messages visible in the conversation array.

## Changes

| File | Change |
|---|---|
| `client-js/rtvi/events.ts` | Add `UserLlmText = "userLlmText"` to `RTVIEvent`; add `userLlmText` to `RTVIEvents` type |
| `client-js/client/client.ts` | Add `onUserLlmText` callback; add `case USER_LLM_TEXT:` in `handleMessage` |
| `client-react/src/conversation/conversationActions.ts` | Add `!lastMerged.final &&` to `shouldMerge` |
| `client-react/src/conversation/useConversationEventWiring.ts` | Wire `RTVIEvent.UserLlmText` → `upsertUserTranscript(..., true)` |

## Test plan

- [ ] Two consecutive bot turns within 30 seconds produce separate bubbles (not merged)
- [ ] In chat mode, user messages appear in the conversation array via `usePipecatConversation`
- [ ] In voice mode, `user-llm-text` still fires and is now surfaced as `RTVIEvent.UserLlmText`
- [ ] Existing `UserTranscript` behavior is unchanged
- [ ] TypeScript builds cleanly in both packages (`tsc --noEmit`)

### DIFF
--- a/client-js/client/client.ts
+++ b/client-js/client/client.ts
@@ -38,6 +38,7 @@ import {
   TranscriptData,
   TransportState,
   UICancelJobGroupData,
+  UserLLMTextData,
   UICommandData,
   UIEventData,
   UIJobGroupData,
@@ -149,6 +150,7 @@ export type RTVIEventCallbacks = Partial<{
   onUserMuteStarted: () => void;
   onUserMuteStopped: () => void;
   onUserTranscript: (data: TranscriptData) => void;
+  onUserLlmText: (data: UserLLMTextData) => void;
   onBotOutput: (data: BotOutputData) => void;
   /** @deprecated Use onBotOutput instead */
   onBotTranscript: (data: BotLLMTextData) => void;
@@ -1142,6 +1144,12 @@ export class PipecatClient extends RTVIEventEmitter {
       case RTVIMessageType.USER_TRANSCRIPTION: {
         const TranscriptData = ev.data as TranscriptData;
         this._options.callbacks?.onUserTranscript?.(TranscriptData);
+        break;
+      }
+      case RTVIMessageType.USER_LLM_TEXT: {
+        const llmTextData = ev.data as UserLLMTextData;
+        this._options.callbacks?.onUserLlmText?.(llmTextData);
+        this.emit(RTVIEvent.UserLlmText, llmTextData);
         break;
       }
       case RTVIMessageType.BOT_OUTPUT: {
--- a/client-js/rtvi/messages.ts
+++ b/client-js/rtvi/messages.ts
@@ -153,6 +153,10 @@ export type BotLLMTextData = {
   text: string;
 };
 
+export type UserLLMTextData = {
+  text: string;
+};
+
 export type BotTTSTextData = {
   text: string;
 };
--- a/client-js/rtvi/events.ts
+++ b/client-js/rtvi/events.ts
@@ -21,6 +21,7 @@ import {
   TranscriptData,
   UICommandData,
   UIJobGroupData,
+  UserLLMTextData,
 } from "./messages";
 
 export enum RTVIEvent {
@@ -65,6 +66,7 @@ export enum RTVIEvent {
   BotTranscript = "botTranscript",
 
   // llm events
+  UserLlmText = "userLlmText",
   BotLlmText = "botLlmText",
   BotLlmStarted = "botLlmStarted",
   BotLlmStopped = "botLlmStopped",
@@ -150,6 +152,7 @@ export type RTVIEvents = Partial<{
   botTranscript: (data: BotLLMTextData) => void;
 
   // llm events
+  userLlmText: (data: UserLLMTextData) => void;
   botLlmText: (data: BotLLMTextData) => void;
   botLlmStarted: () => void;
   botLlmStopped: () => void;
--- a/client-react/src/conversation/conversationActions.ts
+++ b/client-react/src/conversation/conversationActions.ts
@@ -130,6 +130,7 @@ export const mergeMessages = (
       lastMerged.role === currentMessage.role &&
       currentMessage.role !== "system" &&
       currentMessage.role !== "function_call" &&
+      !lastMerged.final &&
       timeDiff < MERGE_WINDOW_MS;
 
     if (shouldMerge) {


client-js/client/client.ts
```diff
@@ -1144,6 +1145,12 @@ export class PipecatClient extends RTVIEventEmitter {
         this._options.callbacks?.onUserTranscript?.(TranscriptData);
         break;
       }
+      case RTVIMessageType.USER_LLM_TEXT: {
+        const llmTextData = ev.data as TranscriptData;
```
mattieruth: the data in a user-llm-text events does not match `TranscriptData`. It simply includes a `text` entry (much like `bot-llm-text`)

client-react/src/conversation/useConversationEventWiring.ts
```diff
@@ -309,6 +309,18 @@ export function useConversationEventWiring() {
     )
   );
 
+  // Chat mode: server emits user-llm-text instead of user-transcription when text
+  // input bypasses STT. Treat it as a finalized user message.
+  useRTVIClientEvent(
```
mattieruth: i haven't had a chance to test this, but I do not think this is what we want. `user-llm-text` will fire for EVERY user entry to the llm, spoken or not. If I'm reading this right, the result is that we will duplicate the user's spoken text. Even if not, I'm not sure this is the correct approach to filling in the missing user text. cc @Regaddi for thoughts.

client-js/client/client.ts
rahulsolanki001: You're right — fixed in the follow-up commit. Defined a proper `UserLlmTextData = { text: string }` type in `messages.ts` and used it throughout (callback declaration, `RTVIEvents` map, `handleMessage` cast). `TranscriptData` carries `final`, `timestamp`, and `user_id` which aren't part of the `user-llm-text` payload.

client-react/src/conversation/useConversationEventWiring.ts
rahulsolanki001: Agreed and reverted in the follow-up commit. You're right that the wiring would cause duplication in voice mode. The exact path:

1. `user-transcription` fires with `final: true` → `upsertUserTranscript` updates the last part to `final: true`
2. `user-llm-text` fires → `upsertUserTranscript` checks `messages[lastUserIndex].final` — still `false` at the message level (finalization only happens on next `UserStartedSpeaking`) — and `lastPart.final === true`, so it **pushes a new part** with the same text. The user message ends up with the transcript text twice.

Removed the `useConversationEventWiring` change entirely. `RTVIEvent.UserLlmText` and `onUserLlmText` are still exposed so callers can handle the event themselves, but we leave proper chat-mode conversation integration for a follow-up that can avoid this duplication path (e.g. an opt-in flag or a dedicated action that checks whether a `user-transcription` already exists for the current turn).

client-js/rtvi/messages.ts
```diff
@@ -153,6 +153,10 @@ export type BotLLMTextData = {
   text: string;
 };
 
+export type UserLlmTextData = {
```
mattieruth: I hate writing nits to people i don't know, but this should be `UserLLMTextData` for consistency


```suggestion
export type UserLLMTextData = {
```

mattieruth (commented):
This is great. and a good catch. Before I approve, I'd like @Regaddi to weigh in on the client-react change.

client-js/rtvi/messages.ts
rahulsolanki001: Good catch, applied — renamed to `UserLLMTextData` everywhere to match `BotLLMTextData`.

Regaddi:
I think the client-react change is fine: it avoids separate messages being merged into a single text bubble.
The branch only needs a proper rebase to resolve the conflicts.

rahulsolanki001:
Rebased onto current main in a single squashed commit — the intermediate fixup commits are now gone. Resolved the `UIJobGroupData`/`UICancelJobGroupData` rename that upstream introduced. Branch shows as mergeable.

mattieruth (approved):
feel free to hit the button :)

Regaddi (approved):

rahulsolanki001:
Thanks so much for the thorough review and the quick turnaround — really appreciated! 🙏

On the topic of the intentionally-deferred part (chat-mode user messages in `usePipecatConversation`): would a follow-up PR be welcome? Here's the approach we had in mind to avoid the voice-mode duplication:

```ts
useRTVIClientEvent(
  RTVIEvent.UserLlmText,
  useAtomCallback(
    useCallback((get, set, data) => {
      const messages = get(messagesAtom);
      const lastUserIdx = findLastIndex(
        messages,
        (m: ConversationMessage) => m.role === "user"
      );

      if (lastUserIdx === -1 || messages[lastUserIdx].final) {
        // Chat mode: no in-progress user message — create one from the LLM text
        upsertUserTranscript(get, set, data.text ?? "", true);
      }
      // Finalize in both cases:
      // voice mode → finalizes the existing message from user-transcription
      // chat mode → finalizes the one we just created
      finalizeLastMessage(get, set, "user");
    }, [])
  )
);
```

The key insight: instead of unconditionally calling `upsertUserTranscript` (which caused the duplicate-part bug), we check whether an in-progress user message already exists. If it does (voice mode — `user-transcription` already handled it), we skip straight to `finalizeLastMessage`. If not (chat mode), we create the message first, then finalize.

Happy to open a PR if this direction looks right to you both!

Regaddi:
@rahulsolanki001 Feel free to open a follow-up PR. Thank you!

rahulsolanki001:
Opened the follow-up as #225 — wires `RTVIEvent.UserLlmText` into `usePipecatConversation` with the voice/chat mode guard we discussed. It's branched from this fix, so it'll need a rebase onto main once this one is merged.

mattieruth:
i'm going ahead and merging this one in the meantime

PR #221, opened 2026-06-01
New callback for handling unsupported features

filipi87:
## Summary
  
  - Added `onUnsupportedFeature` callback and `RTVIEvent.UnsupportedFeature` event to surface transport capability gaps without crashing the application
  - Some transports (e.g. `websocket-transport`) throw `UnsupportedFeatureError` for features like `selectedCam`, `enableCam`, and `enableScreenShare` that they don't implement — previously these propagated as unhandled exceptions
  - `PipecatClient` now catches `UnsupportedFeatureError` in `selectedCam`, `enableCam`, and `enableScreenShare`, routes it through the new event/callback, and returns gracefully instead of throwing 
  
  ## Breaking Changes
  
  None — `onUnsupportedFeature` is an optional entry in the existing `Partial<RTVIEventCallbacks>` type, and `RTVIEvent.UnsupportedFeature` is a new
  enum value. All existing code continues to work unchanged.
  
  ## Testing
  
  Use any transport that doesn't support camera or screen share (e.g. `websocket-transport`) and verify that calling `enableCam()`, `enableScreenShare()`, or reading `selectedCam` no longer throws. Subscribe to the new event to confirm it fires:
  
  ```ts
  client.on(RTVIEvent.UnsupportedFeature, (error) => {
    console.warn("Feature not supported:", error.feature, error.message);
  });

### DIFF
--- a/client-js/client/client.ts
+++ b/client-js/client/client.ts
@@ -898,7 +898,6 @@ export class PipecatClient extends RTVIEventEmitter {
     } catch (e) {
       if (e instanceof RTVIErrors.UnsupportedFeatureError) {
         this._options.callbacks?.onUnsupportedFeature?.(e);
-        this.emit(RTVIEvent.UnsupportedFeature, e);
         return {};
       }
       throw e;
@@ -935,7 +934,6 @@ export class PipecatClient extends RTVIEventEmitter {
     } catch (e) {
       if (e instanceof RTVIErrors.UnsupportedFeatureError) {
         this._options.callbacks?.onUnsupportedFeature?.(e);
-        this.emit(RTVIEvent.UnsupportedFeature, e);
       } else {
         throw e;
       }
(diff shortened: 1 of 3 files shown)


client-js/client/client.ts
```diff
@@ -888,7 +893,16 @@ export class PipecatClient extends RTVIEventEmitter {
   }
 
   public get selectedCam() {
-    return this._transport.selectedCam;
+    try {
+      return this._transport.selectedCam;
+    } catch (e) {
+      if (e instanceof RTVIErrors.UnsupportedFeatureError) {
+        this._options.callbacks?.onUnsupportedFeature?.(e);
+        this.emit(RTVIEvent.UnsupportedFeature, e);
```
mattieruth: isn't this unnecessary? shouldn't the line above call the wrapped callback and emit the error for you? (same goes for `enableCam` and `enableScreenShare`)

filipi87: Yeah, you are right. Inside this._options we are overriding the callbacks by the `wrappedCallbacks`, which internally does both the this.emit.

Nice catch!

filipi87: Fixed!

mattieruth (approved):

PR #228, opened 2026-06-12
Adding support to RTVI protocol 2.0.0

filipi87:
## client-react changes
  
  - Detects the bot's protocol version on `BotReady` and routes `BotOutput` events through either a **legacy path** (Protocol 1.4.x,   client-side word-matching cursor) or a new **v2 path** (Protocol 2.0.0, server-provided `spoken_progress`)
  - Adds `applySpokenProgressV2` to `botOutput.ts`: sets the cursor directly from `accumulated_text.length` using `segment_id` to target the correct part — no word-matching needed
  - Adds a `needsSeparator` flag to `ConversationMessagePart` so the renderer can inject an inter-segment space at display time without modifying the stored segment text 
  - Adds `botOutputProtocolAtom` to track the active protocol version across the session
  - All existing legacy behaviour is preserved and remains the default for bots that haven't upgraded
  
  ## Breaking Changes
  
  - `updateAssistantBotOutput` in `conversationActions.ts` now takes a `BotOutputPayload` discriminated union (`{ protocol: "legacy", spoken }` | `{ protocol: "v2", ... }`) instead of a plain `spoken: boolean` argument — affects any code calling this internal function directly

### DIFF
--- a/client-react/src/conversation/conversationActions.ts
+++ b/client-react/src/conversation/conversationActions.ts
@@ -591,7 +591,7 @@ export function updateAssistantBotOutput(
 
   if (payload.protocol === "v2") {
     // -------------------------------------------------------------------------
-    // RTVI Protocol 2.0.0 path
+    // RTVI Protocol 2.x.x path
     // -------------------------------------------------------------------------
     const { will_be_spoken, spoken_status, spoken_progress, segment_id } =
       payload;
--- a/client-react/src/conversation/conversationActions.ts
+++ b/client-react/src/conversation/conversationActions.ts
@@ -9,6 +9,7 @@ import type { ReactNode } from "react";
 
 import {
   applySpokenBotOutputProgress,
+  applySpokenProgressV2,
   hasUnspokenContent,
   normalizeForMatching,
 } from "./botOutput";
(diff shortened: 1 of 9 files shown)


client-react/src/conversation/conversationActions.ts
```diff
@@ -467,18 +468,42 @@ export function upsertUserTranscript(
...
 ) {
-  // Strip turn-completion markers emitted by FilterIncompleteTurns before
-  // the text enters conversation state.
-  text = stripTurnCompletionMarker(text);
-  if (text.length === 0) return;
+  // v2 progress events carry no text — they only advance the cursor.
+  // Skip text stripping and the empty-text guard for those.
+  const isV2ProgressEvent =
+    payload.protocol === "v2" &&
+    (payload.spoken_status === "in-progress" || payload.spoken_status === "completed");
```
Regaddi: Question: Do we need the checks on `spoken_status` being `"in-progress"` or `"completed"`? What happens to events where `spoken_status` is `"new"`? Aren't these not considered as v2 events?

client-react/src/conversation/conversationActions.ts
```diff
@@ -552,211 +589,268 @@ export function updateAssistantBotOutput(
...
-      const separator =
-        lastPartText && !lastPartText.endsWith(" ") && !text.startsWith(" ")
-          ? " "
-          : "";
-      parts[parts.length - 1] = {
-        ...lastPart,
-        text: lastPartText + separator + text,
+  if (payload.protocol === "v2") {
+    // -------------------------------------------------------------------------
+    // RTVI Protocol 2.0.0 path
```
Regaddi:
```suggestion
    // RTVI Protocol 2.x.x path
```

client-react/src/conversation/conversationActions.ts
filipi87: You already had this code here to strip the turn completion marker. The check above is just an optimization: since we don't use "in_progress" or "completed" events to generate new text, we can skip the code below. It should be safe to remove this check, though, as that's the only reason it's there.
```
text = stripTurnCompletionMarker(text);
if (text.length === 0) return;
```

I will do that.

filipi87: Done!

Regaddi (approved):

PR #235, opened 2026-06-24
Add DTMF support to client-js and client-react, with key-sequence sending

Regaddi:
## Summary
- **client-js**: new `DTMFButton` type (`0`–`9`, `*`, `#`) and a `sendDTMF(dtmf)` method on `PipecatClient` that accepts a single key or a whole key sequence (e.g. `"123#"`). Invalid characters throw an `RTVIError`.
- **client-react**: new `useDTMF()` hook returning `{ sendTone }`, following the existing sender-hook patterns; `sendTone` accepts the same single key or sequence.
- **Wire format / compatibility**: the client tracks the bot's RTVI protocol version from `bot-ready`. Bots on protocol 2.1.0+ (pipecat-ai/pipecat#5030) receive one `{ type: "dtmf", data: { buttons: ["1", "2", "3", "#"] } }` message; 2.0.x bots (pipecat 1.5.x) receive one legacy `{ button }` message per key, so DTMF works against every deployed server that supports it. Bots older than protocol 2.0.0 (or with no advertised version) predate the `dtmf` message entirely, so `sendDTMF` throws `UnsupportedFeatureError` instead of silently sending messages they'd reject. `RTVI_PROTOCOL_VERSION` is bumped to `2.1.0`.

## Test plan
- [x] Jest suites pass for both packages, covering: a sequence as a single `buttons` message to 2.1.0+ bots, a single key as a one-element `buttons` message, per-key legacy fan-out for 2.0.x bots, `UnsupportedFeatureError` for pre-2.0.0/unknown bot versions, and invalid input throwing without sending
- [x] `tsc --noEmit` and `eslint` pass for both packages
- [ ] Manual: against a pipecat bot with a `DTMFAggregator`, call `client.sendDTMF("123#")` and confirm the keys aggregate into one transcription
- [ ] Manual: `useDTMF().sendTone("#")` from a React component dispatches the same message

### DIFF
--- a/client-js/client/client.ts
+++ b/client-js/client/client.ts
@@ -18,6 +18,8 @@ import {
   ClientMessageData,
   DeviceErrorReason,
   DeviceStatus,
+  DTMFButton,
+  DTMFData,
   ErrorData,
   LLMContextMessage,
   LLMFunctionCallData,
@@ -39,11 +41,11 @@ import {
   TranscriptData,
   TransportState,
   UICancelJobGroupData,
-  UserLLMTextData,
   UICommandData,
   UIEventData,
   UIJobGroupData,
   UISnapshotData,
+  UserLLMTextData,
 } from "../rtvi";
 import * as RTVIErrors from "../rtvi/errors";
 import {
@@ -996,6 +998,18 @@ export class PipecatClient extends RTVIEventEmitter {
   public sendUIEvent(event: string, payload?: unknown): void {
     const data: UIEventData = { event, payload };
     this._sendMessage(new RTVIMessage(RTVIMessageType.UI_EVENT, data));
+  }
+
+  /**
+   * Send a single DTMF tone to the server as a first-class RTVI
+   * `dtmf` message. Call once per key.
+   *
+   * @param button - The DTMF key to send (0-9, *, or #).
+   */
+  @transportReady
+  public sendDTMF(button: DTMFButton): void {
+    const data: DTMFData = { button };
+    this._sendMessage(new RTVIMessage(RTVIMessageType.DTMF, data));
   }
 
   /**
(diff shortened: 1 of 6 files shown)


markbackman (approved):
LGTM!

client-js/client/client.ts
```diff
@@ -998,6 +1004,36 @@ export class PipecatClient extends RTVIEventEmitter {
...
+  @transportReady
+  public sendDTMF(dtmf: DTMFButton | string): void {
+    if (!/^[0-9*#]+$/.test(dtmf)) {
+      throw new RTVIErrors.RTVIError(
+        `Invalid DTMF sequence "${dtmf}". Only 0-9, * and # are allowed.`
+      );
+    }
+    const buttons = [...dtmf] as DTMFButton[];
+    const botSupportsBatch =
+      this._botVersion[0] > 2 ||
```
markbackman: Including this version check in case someone is using pipecat-ai 1.5.0 which was RTVI 2.0.0, which had a different type (button: KeypadEntry).

mattieruth: I think this should be:
```
// check if supports DTMF at all
if (_botVersion[0] >= 2) {
  if (_botVersion[0] === 2 && _botVersion[1] < 1) {
     iterate and send with button
  } else {
    send buttons
  }
} else {
  ERROR
}
```

mattieruth (approved):
just the one thing:

client-js/client/client.ts
markbackman: Agreed — the old fallback conflated "bot that accepts per-key DTMF" (2.0.x) with "bot that doesn't do DTMF at all" (< 2.0.0 / unknown), and fanning out to the latter just generates server-side 'Unsupported type' errors the caller never sees. Restructured as you suggested (folded into 5c5ce54), using the existing `UnsupportedFeatureError` for the error branch. Timing-wise it can't fire spuriously: `_botVersion` is set in the BOT_READY handler before `connect()` resolves, so an unknown version genuinely means a bot that never advertised 2.x.

Issue #223, opened 2026-06-04
client-react inlines a private copy of client-js types (RTVIEvent, …), breaking useRTVIClientEvent typings across versions


sparanoid:
### Summary

`@pipecat-ai/client-react`'s published types (`dist/index.d.ts`) **inline a private copy** of `@pipecat-ai/client-js`'s type surface — the `RTVIEvent` enum plus ~70 other declarations — instead of importing them from `@pipecat-ai/client-js`. Because TypeScript enums are **nominal**, that inlined `RTVIEvent` is a *different type* from the `RTVIEvent` consumers import from `@pipecat-ai/client-js`. As soon as the installed `client-js` diverges from the copy a given `client-react` was built against, every `useRTVIClientEvent(RTVIEvent.X, …)` call fails to type-check.

### Reproduction

Install `@pipecat-ai/client-js@1.11.0` + `@pipecat-ai/client-react@1.6.0`, then:

```ts
import { RTVIEvent } from '@pipecat-ai/client-js'
import { useRTVIClientEvent } from '@pipecat-ai/client-react'

useRTVIClientEvent(RTVIEvent.BotReady, () => {})
```

`tsc --noEmit`:

```
error TS2345: Argument of type 'RTVIEvent.BotReady' is not assignable to parameter of type 'RTVIEvent'.
```

This reproduces for every event member. It surfaced specifically with `client-js@1.11.0` because #221 added `RTVIEvent.UnsupportedFeature`, widening the enum beyond the copy frozen into `client-react@1.6.0` (built against an earlier `client-js`).

### Root cause

`client-react/tsconfig.json` aliases the dependency to client-js's **local source**:

```jsonc
"paths": {
  "@/*": ["./src/*"],
  "@pipecat-ai/client-js": ["../client-js/index.ts"]
}
```

Parcel's `@parcel/transformer-typescript-types` follows that alias into source and **inlines** client-js's declarations into `client-react/dist/index.d.ts` instead of emitting `import { … } from '@pipecat-ai/client-js'`. You can confirm it in the shipped file: `enum RTVIEvent { … }` is declared locally, and `client-react` never re-exports it — so consumers can only obtain `RTVIEvent` from `client-js`, guaranteeing the nominal mismatch.

### Impact

[...]

PR #239, opened 2026-07-15
fix(client-react): stop bundling client-js types into the published d.ts

Regaddi:
The `@pipecat-ai/client-js` -> `../client-js/index.ts` path alias made Parcel's TypeScript type bundler resolve client-js to local source and inline it into `dist/index.d.ts`. As a result the published types declared their own duplicate `PipecatClient` class (referencing a bare `"client-js"` module) instead of re-exporting the one from `@pipecat-ai/client-js`.

That duplicate is nominally distinct from the real client-js `PipecatClient` (protected members), so a client-js `PipecatClient` instance is not assignable to `PipecatClientProvider`'s `client` prop. Any TypeScript consumer that creates a `PipecatClient` and passes it to the provider fails to type-check (and downstream `.d.ts` builds break).

Removing the alias lets the type bundler treat `@pipecat-ai/client-js` as an external dependency again, so the emitted `dist/index.d.ts` imports `PipecatClient`, `DTMFButton`, etc. from `@pipecat-ai/client-js` with no inlined duplicate. The runtime bundle already externalized client-js correctly; only the type build was affected.

Note: client-react now type-checks against client-js's built declarations, so build client-js before client-react (the workspace build order already does this).

### DIFF
--- a/client-react/tsconfig.json
+++ b/client-react/tsconfig.json
@@ -18,8 +18,7 @@
     /* Path aliases */
     "baseUrl": ".",
     "paths": {
-      "@/*": ["./src/*"],
-      "@pipecat-ai/client-js": ["../client-js/index.ts"]
+      "@/*": ["./src/*"]
     },
 
     /* Linting */


markbackman (approved):

gregauger:
This duplicates issue https://github.com/pipecat-ai/pipecat-client-web/issues/223 and PR https://github.com/pipecat-ai/pipecat-client-web/pull/224 (where the fix is identical to this PR). You can close them now.

Glad it has been fixed, but a bit confused as to why it has taken so long / the other PR wasn't looked at.

mattieruth:
> This duplicates issue #223 and PR #224 (where the fix is identical to this PR). You can close them now.
> 
> Glad it has been fixed, but a bit confused as to why it has taken so long / the other PR wasn't looked at.

TBH, I think this one just kept falling through the cracks such that when we finally did it, we lost sight that there was already a PR open. Sorry about that!

PR #242, opened 2026-07-28
- Version gate checks protocol < 2.2 (was < 1.3, which let 2.0/2.1 bo…

vipyne:
…ts through).

- FileReader path strips the data-URL prefix so bytes carries raw base64, uses the normalized MIME type, includes the filename, and properly rejects on read errors (previously an unreadable file hung the promise forever).
- sendFile no longer mutates the caller's RTVIFile (spreads a copy).
- messages.ts: "hief" → "heif", removed the duplicate "ogg". Rebuilt client-js — your running example picks up the new dist on refresh, and since the local pipecat branch now advertises 2.2.0, local end-to-end still works.

### DIFF
--- a/client-js/client/client.ts
+++ b/client-js/client/client.ts
@@ -1168,8 +1168,8 @@ export class PipecatClient extends RTVIEventEmitter {
     options: SendFileOptions = {}
   ) {
     if (
-      this._botVersion[0] < 1 ||
-      (this._botVersion[0] === 1 && this._botVersion[1] < 3)
+      this._botVersion[0] < 2 ||
+      (this._botVersion[0] === 2 && this._botVersion[1] < 2)
     ) {
       throw new RTVIErrors.UnsupportedFeatureError(
         "sendFile",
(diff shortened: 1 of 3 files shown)


client-js/client/client.ts
```diff
@@ -1205,23 +1205,33 @@ export class PipecatClient extends RTVIEventEmitter {
...
           reader.onload = async (e) => {
-            if (!e.target?.result) {
-              throw new RTVIErrors.RTVIError("Could not read file data");
+            try {
+              if (!e.target?.result) {
+                throw new RTVIErrors.RTVIError("Could not read file data");
+              }
+              const dataUrl = e.target.result as string;
+              // FileBytes.bytes carries raw base64; strip the data-URL prefix.
+              const base64Data = dataUrl.split(",")[1] ?? dataUrl;
```
mattieruth, line 1220: i think you're right here, but i swear i tested this. 😕

client-js/rtvi/messages.ts
```diff
@@ -278,7 +278,7 @@ export type RTVIImageFormat =
   | "webp"
   | "gif"
   | "heic"
-  | "hief";
+  | "heif";
```
mattieruth, line 281: 🤦‍♀️

PR #243, opened 2026-08-11
fix(client-react): Fixed issue where mic/cam enabled state could get …

mattieruth:
…out of sync

the mic/cam can turn off for more than the reason that they were manually turned off via enableCam/Mic, like when the client disconnects or if there's an error. This fix listens for track-started/stopped events to ensure that the provider state remains in sync with the underlying mic/cam state.

### DIFF
--- a/client-react/src/PipecatClientState.tsx
+++ b/client-react/src/PipecatClientState.tsx
@@ -100,6 +100,25 @@ export const PipecatClientStateProvider: React.FC<React.PropsWithChildren> = ({
     }
   });
 
+  // TrackStarted/Stopped fire for more than an enabled/disabled toggle — a
+  // device switch is Stopped(old)+Started(new) with the enabled state never
+  // conceptually changing, so treating "stopped" as false / "started" as
+  // true would drift from reality whenever those causes interleave. Use the
+  // event purely as a signal to re-check, and read the actual state off the
+  // client (the source of truth) rather than inferring it from which event
+  // fired.
+  useRTVIClientEvent(RTVIEvent.TrackStarted, (track, participant) => {
+    if (!participant?.local || !client) return;
+    if (track.kind === "audio") setIsMicEnabled(client.isMicEnabled ?? false);
+    if (track.kind === "video") setIsCamEnabled(client.isCamEnabled ?? false);
+  });
+
+  useRTVIClientEvent(RTVIEvent.TrackStopped, (track, participant) => {
+    if (!participant?.local || !client) return;
+    if (track.kind === "audio") setIsMicEnabled(client.isMicEnabled ?? false);
+    if (track.kind === "video") setIsCamEnabled(client.isCamEnabled ?? false);
+  });
+
   return (
     <PipecatClientTransportStateContext.Provider value={transportState}>
       <PipecatClientCamStateContext.Provider


Regaddi:
I think Claude isn't entirely right on a couple of the following points. I'll have to manually check. Haven't gotten to it, yet!

---

Approach looks right, and better than the obvious alternative: using the event purely as a re-check trigger and reading `client.isMicEnabled` as the source of truth survives a device switch (Stopped(old)+Started(new)), which a naive `stopped -> false` would break. The comment explaining that is worth keeping.

Verified while reading: screen tracks use separate events so there's no cross-talk; `WavMediaManager.enableMic` sets `_micEnabled` before emitting `onTrackStopped`, so the read-back isn't stale; `SmallWebRTCTransport`'s remote-track handlers pass no participant, so the `participant?.local` guard correctly skips them; and `?? false` is load-bearing since `daily.localAudio()` returns `null` when not in a call.

A few things:

**1. Does this fix the "error" half?** I can't verify from inside this repo, so flagging rather than asserting. For daily-backed transports `isMicEnabled` resolves to `daily.localAudio()`. If that reflects *intended* mute state rather than track liveness, then in the failure case the description calls out (device unplugged, permission revoked, OS grabs the mic) the track ends, `TrackStopped` fires, but `localAudio()` may still return `true` and the state drifts exactly as before. The disconnect case looks genuinely fixed. Which scenario did you reproduce, and is `MediaStateUpdated` / `client.mediaState` a better signal for the error half?

**2. The optimistic setState is still a drift source, in the same scope.** `enableCam` sets state to `true` first, then calls `client.enableCam()`, which swallows `UnsupportedFeatureError`. No track event ever fires, so the provider reports the cam as on permanently. Same shape for `enableMic` if the transport rejects. This PR closes "something else turned it off" but leaves "it never turned on" open.

**3. The two handlers are byte-identical.** Worth collapsing into one `useCallback` passed to both. That also removes churn: `useRTVIClientEvent` lists `handler` in its effect deps, so inline arrows re-subscribe on every provider render, and this provider re-renders on every state change.

**4. No test.** `client-react` already has jest + jsdom + `@testing-library/react`, with `useDTMF.spec.tsx` as precedent. A test asserting that local-audio `TrackStopped` re-reads `client.isMicEnabled` would pin a regression whose correctness depends entirely on non-obvious ordering.

**5. Trivial:** the PR title scope `client-react` isn't in `.commitlintrc.json`'s scope-enum (`js`, `react`, empty). The commit itself is correctly `fix(react):`; only bites if this is squash-merged with the title as the subject.

Regaddi:
Followed up on my Q1 by reading the transports repo instead of guessing. Resolving it:

`isMicEnabled` bottoms out at daily-js `localAudio()`, which is:

```js
this._participants.local
  ? !["blocked", "off"].includes(this._participants.local.tracks.audio.state)
  : null
```

(Both `DailyTransport` and small-webrtc's `DailyMediaManager` use this; small-webrtc's `WavMediaManager` uses an intended-state `_micEnabled` instead.) And daily-js fires `track-stopped` on the transition *out of* `"playable"`.

So mapping the causes the PR names:

- **Disconnect** → local participant gone → `localAudio()` returns `null` → `?? false` → synced. ✅
- **User mute / permission block / device-missing** → state `off` / `blocked` → `localAudio()` false → synced. ✅
- **Track `interrupted`** (OS grabs the mic, transient device interruption) → `track-stopped` *does* fire, but `"interrupted"` isn't in the `["blocked","off"]` exclusion, so `localAudio()` still returns `true`. The handler re-reads `true` and the UI keeps showing the mic on. ⚠️

So the fix genuinely covers disconnect and the mute/block cases (the common ones), but the `interrupted` track state is still a drift in the same direction as the original bug. Probably fine to merge as-is if that case is out of scope, worth a comment noting it if not. The other points (optimistic-setState hole on `enableCam`/`enableMic` reject, dedup the two handlers, test) still stand.

Regaddi (approved):

mattieruth:
of note: for the interrupted case, the current behavior is what we want. `micEnabled`/`camEnabled` is meant to represent the intended state. blocked or off is an intentional or at least legitimate mute. interrupted is likely temporary or going against the intended state. so i think we're good on all fronts.

PR #247, opened 2026-08-27
docs: add the update-docs profile for pipecat-client-web

markbackman:
Groundwork for giving this repo the documentation automation `pipecat-ai/pipecat` has. Companion to [daily-co/pipecat-cloud#201](https://github.com/daily-co/pipecat-cloud/pull/201).

The shared `update-docs` skill lives in `pipecat-ai/pipecat` and is published through the `pipecat-dev-skills` marketplace. It supplies the workflow; each documented repo supplies a **profile** mapping its source to the pages that document it. This is that half.

## Two things make this shaped differently from the Python profiles

**The mapping isn't one file to one page.** Sixteen `use*.ts` hooks share `react/hooks.mdx`. `PipecatClient` splits across three pages — `PipecatClientOptions` to the constructor page, methods to `client-methods.mdx`, `RTVIEventCallbacks` to `callbacks.mdx`. Resolving a change means finding the *section*, and a new hook or component is a new **section**, not a new page. Both rules are stated explicitly, since an agent applying Python-shaped intuition would create fifteen pages nobody wants.

**This repo doesn't own every page under `api-reference/client/js/`.** The six concrete transports are published from `pipecat-client-web-transports` and documented in the same directory. Only `transports/transport.mdx` — the base class — belongs here. The profile says to report changes affecting the others as cross-repo findings rather than editing another package's docs from source that doesn't define it. Nine pages owned here, six owned there.

## Contract checks

- **Forward:** all 55 in-scope source files reach a rule.
- **Backwards:** all 9 pages this repo owns are reachable by one.

The backwards check earned its keep again — and the forward one caught something I'd got outright wrong. I had written that the UI and a11y surface (`rtvi/ui.ts`, `a11y_walker.ts`, `A11ySnapshotStreamer.ts`) has *no page*. It does: it's documented through the React hooks that consume it — `useUISnapshot`, `useUICommandHandler`, `useUIJobGroups`. Left as written, every change to that surface would have been reported as a documentation gap that isn't one. Three more files (`common_types.ts`, `defaultUICommandHandlers.ts`, `uiJobGroupsTypes.ts`) were unmapped for the same reason.

## One deliberate note in the Scope section

Barrel files (`index.ts`) are excluded because they re-export — but if one **stops** exporting a name, that name has left the public API, which is a documentation change. The profile says to treat a removed export line as a change to whatever page documents that name, so a removal isn't invisible just because the file is skipped.

## Not included: the workflow

Deliberate. `update-docs.yml` should copy whatever the end-to-end dispatch test proves out on pipecat, rather than being written speculatively — and the version in `pipecat-flows` is the pre-fix one (`max-turns 30`, no formatting step) that shouldn't be propagated.

### DIFF
--- a/.claude/skills/update-docs/SOURCE_DOC_MAPPING.md
+++ b/.claude/skills/update-docs/SOURCE_DOC_MAPPING.md
@@ -1,0 +1,168 @@
+# Source-to-Doc Mapping — pipecat-client-web
+
+The profile for the shared `update-docs` skill, which lives in
+`pipecat-ai/pipecat` at `.claude/skills/update-docs/SKILL.md` and is published
+through the `pipecat-dev-skills` marketplace. `PROFILE_CONTRACT.md` beside it
+describes what this file has to provide.
+
+Two differences from the Python profiles are worth reading before using the
+tables below.
+
+**The mapping is not one file to one page.** Sixteen `use*.ts` hooks share a
+single page, and one class (`PipecatClient`) is split across three. Resolving a
+change means finding the *section* it belongs to, not just the file.
+
+**This repo does not own every page under `api-reference/client/js/`.** The six
+concrete transports are published from `pipecat-ai/pipecat-client-web-transports`
+and documented in the same directory. Only `transports/transport.mdx` — the base
+class — belongs to this repo. Editing a concrete transport's page from here
+means editing another package's documentation from source that does not define
+it.
+
+## Scope
+
+Everything under `client-js/` and `client-react/src/` is in scope. Both are
+published packages (`@pipecat-ai/client-js`, `@pipecat-ai/client-react`).
+
+Exclude:
+
+- `client-js/tests/**`, `client-react/tests/**`
+- `node_modules/`, `dist/`, `*.d.ts` build output
+- `client-js/index.ts`, `client-js/client/index.ts`, `client-js/rtvi/index.ts`,
+  `client-react/src/index.ts` — barrel files that only re-export
+
+A barrel file is excluded because it re-exports, not because it is uninteresting:
+if one **stops** exporting a name, that name has left the public API, which is a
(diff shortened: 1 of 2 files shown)


.claude/skills/update-docs/SOURCE_DOC_MAPPING.md
```diff
@@ -0,0 +1,168 @@
...
+| `client-js/rtvi/events.ts` | `api-reference/client/js/callbacks.mdx` — `RTVIEvent` is the event-name enum the callbacks table is keyed on |
+| `client-js/rtvi/errors.ts` | `api-reference/client/js/errors.mdx` |
+| `client-js/rtvi/common_types.ts` | Wherever the changed type is used. `TransportState` appears on `api-reference/client/js/callbacks.mdx`, `js/transports/transport.mdx`, and `react/hooks.mdx`; `Participant` on `callbacks.mdx`. Grep the type name — these are referenced far more widely than they are defined. |
+| `client-js/rtvi/ui.ts`, `client-js/rtvi/a11y_walker.ts`, `client-js/client/A11ySnapshotStreamer.ts` | `api-reference/client/react/hooks.mdx` — the UI-command and snapshot surface is documented through the React hooks that consume it (`useUISnapshot`, `useUICommandHandler`, `useUIJobGroups`), not on a page of its own |
+| `client-js/client/rest_helpers.ts` | `api-reference/client/js/client-constructor.mdx` — the `endpoint`/`requestData` connect helpers |
+| `client-js/client/logger.ts` | `api-reference/client/js/overview.mdx` |
+
+## Pattern matching
+
+Two many-to-one rules cover most of `client-react/`:
```
mattieruth, line 72: i hope robots understand this line. it's a doozy

mattieruth (approved):
Image

PR #248, opened 2026-08-28
ci(docs): open documentation PRs when public API changes

markbackman:
Installs the docs automation this repo has never had. **Stacked on #247** (the profile) — review that first; this PR's diff shows only the workflow once #247 merges.

## Not a rewrite

pipecat's workflow with four adaptations, so everything it has accumulated carries over:

- the **90-turn cap** that fixed a ~40% silent-drop rate
- **formatting before `llms.txt` generation** — the reverse order fails the docs repo's lint
- **outcome recording**, so a run that documented nothing is distinguishable from one that produced a PR
- **failure reporting** back onto the source PR, with a label and a re-run command

| | |
|---|---|
| **Scope** | `client-js/**` and `client-react/src/**`, excluding both tests directories |
| **Branch names** | `docs/client-web-pr-N` |
| **Token scope** | App token for `owner: pipecat-ai`, scoped to `docs` alone |
| **Source repo refs** | every `gh` call targets `pipecat-ai/pipecat-client-web` |

## Ownership note in the header

The workflow carries a comment saying this repo doesn't own every page under `api-reference/client/js/` — the six concrete transport pages belong to `pipecat-client-web-transports`, and only `transports/transport.mdx` is ours. The profile says the same, but someone reading the workflow to see what it touches shouldn't have to open a second file to find out.

## Secrets are already set

`DOCS_BOT_APP_ID`, `DOCS_BOT_PRIVATE_KEY`, and `ANTHROPIC_API_KEY` are provisioned on this repository. The App's credentials were verified against GitHub directly — it authenticates as `pipecat-docs-bot` (id 4080679) and holds `contents: write` and `pull_requests: write` on `pipecat-ai/docs`, which is what pushing a branch and opening a PR there requires.

The App needs no grant on *this* repo: the workflow reads its own PRs with the automatic `GITHUB_TOKEN`.

## Honest caveat

The shared skill has been verified statically — the profile's nine contract sections are present, all 9 owned pages are reachable by a rule, the YAML parses with 12 steps — but has **not been run end to end** since being refactored. The first merged PR here will be its first live exercise, and a failure surfaces as a labelled red run rather than a wrong docs PR.

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -66,6 +66,18 @@ jobs:
           token: ${{ steps.app-token.outputs.token }}
           path: _docs
 
+      # The update-docs skill is shared and lives in pipecat-ai/pipecat, published
+      # through the pipecat-dev-skills marketplace. Only the directory holding it
+      # is fetched; this repo supplies the profile beside it.
+      - name: Checkout the shared update-docs skill
+        uses: actions/checkout@v4
+        with:
+          repository: pipecat-ai/pipecat
+          sparse-checkout: .claude/skills/update-docs
+          sparse-checkout-cone-mode: false
+          path: _skill
+          fetch-depth: 1
+
       - name: Record docs baseline
         id: docs-base
         working-directory: _docs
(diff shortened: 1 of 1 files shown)


mattieruth (approved):
Image

PR #250, opened 2026-08-31
fix(client-js): align ErrorData with server payload

### LINKED ISSUE #244: ErrorData type says "message" but the pipecat python server sends "error"
The TS client's `ErrorData` type does not match what the Python server actually sends for RTVI `error` messages.

**Python (pipecat, current `main` and 1.7.0)** — `processors/frameworks/rtvi/models.py`:

```python
class ErrorData(BaseModel):
    error: str
    fatal: bool
```

sent by `RTVIProcessor._send_error_frame`:

```python
message = RTVI.Error(data=RTVI.ErrorData(error=frame.error, fatal=frame.fatal))
```

**TS client (`@pipecat-ai/client-js`, checked 1.4.1 through latest 1.13.0)** — `dist/index.d.ts`:

```ts
export type ErrorData = {
    message: string;
    fatal: boolean;
};
```

So on the wire the payload is `{"data": {"error": "...", "fatal": false}}`, and any client code that trusts the published type and reads `data.message` gets `undefined`. We hit this in production-adjacent testing: our error relay rendered `[object Object]` and a downstream classifier keyed on the message text silently never matched, because we'd written the handler against `ErrorData` as typed.

Either the Python field should be renamed `message`, or `ErrorData` in the client should be `{ error: string; fatal: boolean }` (or accept both during a deprecation window) — but as it stands the published type actively misleads.

Happy to PR whichever side you'd prefer to change.

swayamg20:
Fixes #244.

## Summary
- align `ErrorData` with the server's canonical `error` field
- preserve the deprecated `message` field on client-generated errors
- cover server and local error payloads

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,7 +1,7 @@
 {
   "extends": ["@commitlint/config-conventional"],
   "rules": {
-    "scope-enum": [2, "always", ["js", "react", ""]],
+    "scope-enum": [2, "always", ["js", "client-js", "react", "client-react", ""]],
     "body-max-line-length": [0],
     "footer-max-line-length": [0],
     "header-case": [0],
--- a/client-js/rtvi/messages.ts
+++ b/client-js/rtvi/messages.ts
@@ -118,8 +118,10 @@ export type ClientReadyData = {
 };
 
 export type ErrorData = {
-  message: string;
+  error: string;
   fatal: boolean;
+  /** @deprecated Use `error` instead. */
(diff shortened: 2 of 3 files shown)


filipi87 (approved):
LGTM!

mattieruth:
fixing the commit message so we can merge...

swayamg20:
Thanks @filipi87 and @mattieruth!

mattieruth:
Thank you!!

